PR #133, opened 2018-03-13
Add a way to transform TypeMap keys

justinvp:
There are cases where we'd like to transform TypeMap keys, like in pulumi-github. This change adds a `MangleTypeMaps` option for doing so.

Fixes #111


lukehoban:
Closing as this is stale now. We'll continue to track the issue with #111.

Issue #173, opened 2018-06-10
Fix Python mangling for CAPS names


joeduffy:
It appears we are inserting `_` characters between all CAPS characters, e.g. `c_a_p_s`, in our Python name mangling.  For instance, see sdk/python/pulumi_gcp/storage/bucket_a_c_l.py.

PR #179, opened 2018-06-13
Fix camel case to snake case converter implementation

swgillespie:
Fixes https://github.com/pulumi/pulumi-terraform/issues/173.

Camel case to snake case conversion is tricky. I tried a couple different Go libraries that claim to do this, and although they all "work", each one has a slightly different idea of how numbers interact with camel case to snake case conversion that resulted in less-than-optimal resource names. In particular, most of the libraries I tried considered numbers to be separate from lower case letters, so things such as `Sha256Hash` got turned into `sha_256_hash`, which resulted in a lot of breaking name changes in the `pulumi-gcp` provider.

In the hopes of being absolutely sure what sort of names that tfgen produces, and being unable to find a library that does exactly what we want, this PR has a camel case to snake case implementation. This implementation considers digits to be identical to lowercase numbers, so that `Sha256Hash` is treated as `sha256_hash`, which matches the implementation today while still fixing #173 .

The implementation itself is a straight-up state machine. Most of the libraries I looked at take some sort of state machine approach, using either regexes or a hand-coded approach. I chose the latter here because I can wrap my head around it a little easier, as much as I love writing insanely complex regular expressions.

### DIFF
--- a/pkg/tfgen/generate_python.go
+++ b/pkg/tfgen/generate_python.go
@@ -675,7 +675,7 @@ func pyName(name string) string {
 	//                was either a number or a lowercase letter.
 	//   stateAcronym - The last character we saw was an uppercase letter and the character before it
 	//                  was an uppercase letter.
-	//   stateLowerOrNumber - The last character we saw was a letter or a number.
+	//   stateLowerOrNumber - The last character we saw was a lowercase letter or a number.
 	//
 	// The following are the state transitions of this state machine:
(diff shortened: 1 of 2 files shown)


swgillespie:
Reviving this post-launch, I think I've addressed all feedback so I'm going to merge in a little bit if there aren't any objections.

swgillespie:
thanks!

PR #182, opened 2018-06-19
Populate config defaults before checking for missing keys.

### LINKED ISSUE #181: Default config not working?
For the Azure provider, the Terraform docs claim that the `environment` config variable will default to `public`.  And yet, if I omit it, I see

```
    error: Missing required configuration variable 'azure:environment'
    	please set a value using the command `pulumi config set azure:environment <value>`
```

Sure enough, running `pulumi config set azure:environment public` unblocks things.

This seems like perhaps a bug in the way we are projecting Terraform configuration?

pgavlin:
The check for missing keys should come after the provider config has
been converted from its Pulumi representatinon to its TF representation.
This give the TF provider a chance to populate any missing properties
with appropriate defaults.

Fixes #181.

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -189,37 +189,8 @@ func (p *Provider) Configure(ctx context.Context, req *pulumirpc.ConfigureReques
 		}
 	}
 
-	// So we can provide better error messages, do a quick scan of required configs for this
-	// schema and report any that haven't been supplied.
-	var missingKeys []*pulumirpc.ConfigureErrorMissingKeys_MissingKey
-	for key, meta := range p.config {
-		_, present := vars[resource.PropertyKey(key)]
-		if meta.Required && !present {
-			fullyQualifiedKey := tokens.NewModuleToken(p.pkg(), tokens.ModuleName(key))
-
-			// TF descriptions often have newlines in inopportune positions. This makes them present
-			// a little better in our console output.
-			descriptionWithoutNewlines := strings.Replace(meta.Description, "\n", " ", -1)
-			missingKeys = append(missingKeys, &pulumirpc.ConfigureErrorMissingKeys_MissingKey{
-				Name:        fullyQualifiedKey.String(),
-				Description: descriptionWithoutNewlines,
-			})
-		}
-	}
-
-	if len(missingKeys) > 0 {
-		err := rpcerror.New(codes.InvalidArgument, "required configuration keys were missing")
-
-		// Clients of our RPC endpoint will be looking for this detail in order to figure out
-		// which keys need descriptive error messages.
-		err = rpcerror.WithDetails(err, &pulumirpc.ConfigureErrorMissingKeys{
-			MissingKeys: missingKeys,
-		})
-
-		return nil, err
-	}
-
-	// Now make a Terraform config map out of the variables.
+	// First make a Terraform config map out of the variables. We do this before checking for missing properties
+	// s.t. we can pull any defaults out of the TF schema.
 	config, err := MakeTerraformConfig(nil, vars, p.config, p.info.Config, true)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not marshal config state")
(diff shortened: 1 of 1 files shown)


pgavlin:
While investigating this, I noticed that we're in sort of an odd place w.r.t. provider vs. program config. Consider the case of `azure:environment`: if this config var is not set, the provider will compute a default value of `"public"` and use that. This default will not be reflected in the program config, though, so the package will still fail to load with a "missing config" error due to a call to `config.require("environment")`.

In general, allowing providers to compute defaults for config vars that are not set risks causing a mismatch between the config observed by the program and the config observed by the provider. One possible solution is to allow the provider to communicate its computed config to the engine. This would be a pretty invasive change: not only would the provider interface would need to change to accommodate the new info, the way the language providers fetch config would also need to change. Currently config is specified to the language provider when it is loaded after which it cannot be modified. We would need to find some reasonable way around this in order to allow a resource provider to write its computed config back.

swgillespie:
> This would be a pretty invasive change: not only would the provider interface would need to change to accommodate the new info, the way the language providers fetch config would also need to change. Currently config is specified to the language provider when it is loaded after which it cannot be modified. We would need to find some reasonable way around this in order to allow a resource provider to write its computed config back.

Haven't read the code yet, but this may not be a bad thing. Every new language that we support has to implement `pulumi.Config`, which pretty much only involves parsing a string from an environment variable and doing some type casts. The engine gives us the config when launching the process, so it's not unreasonable to me to instead have the engine keep the config and allow the language host to ask for it when necessary. Furthermore, I don't think that there's a backcompat hazard here: we can continue to pass config via environment variables while still also providing an engine endpoint for fetching config. Newer SDKs can ask the engine for config and get the correct behavior for default configs, while older SDKs can continue reading the environment variable.

I think that reducing the amount of code that needs to be in language SDKs is goodness in general.

pkg/tfbridge/provider.go
```diff
@@ -187,48 +187,46 @@ func (p *Provider) Configure(ctx context.Context, req *pulumirpc.ConfigureReques
...
+
 	// So we can provide better error messages, do a quick scan of required configs for this
 	// schema and report any that haven't been supplied.
 	var missingKeys []*pulumirpc.ConfigureErrorMissingKeys_MissingKey
 	for key, meta := range p.tf.Schema {
-		_, present := vars[resource.PropertyKey(key)]
-		if meta.Required && !present {
-			fullyQualifiedKey := tokens.NewModuleToken(p.pkg(), tokens.ModuleName(key))
+		if meta.Required && !config.IsSet(key) {
+			name := TerraformToPulumiName(key, meta, false)
```
swgillespie: what's the difference between a terraform and pulumi name here?

pgavlin: Terraform names `look_like_this`; Pulumi names `lookLikeThis`. When we iterate over the TF schema map we get TF names, but we want to report Pulumi names in the error.

swgillespie: gotcha, thanks!

lukehoban:
I like the direction of having language hosts reach back to the engine to ask about config.  @swgillespie's notes on doing this in a backward compatible way are compelling.  This will also presumably be an important stepping stone as part of addressing https://github.com/pulumi/pulumi/issues/1458 to allow providers to be configured based on outputs from other providers/resources.

One short term note though - to address the issue of `config.get(...)` returning the wrong values - does it make sense to allow `resources.go` to re-specify defaults that will get embedded into the config defaults in the generated code?  That would require manual work in the projection layers, but would be a simple way to enable us to fix these if/when we identify them.

pgavlin:
> One short term note though - to address the issue of config.get(...) returning the wrong values - does it make sense to allow resources.go to re-specify defaults that will get embedded into the config defaults in the generated code? That would require manual work in the projection layers, but would be a simple way to enable us to fix these if/when we identify them.

The major issue with this is the fact that computing a default may run arbitrary code, so there's not necessarily a way to understand what the default might be at gen time. All of the Azure provider's defaults are computed in this fashion.

pgavlin:
@lukehoban finally getting back around to this one--sorry for the delay.

RE: my earlier concerns about provider configuration differing from library configuration: we should be in a _fairly_ reasonable place even with these changes, at least for static defaults. `tfgen` should be emitting code that sets the default value in the library, so as long as we're dealing with defaults that are provided as data (i.e. `DefaultValue`) rather than code (i.e. `DefaultValueFunc`), things should look pretty good. Unfortunately, this does fall apart for exactly the default mentioned in #181: the `environment` variable in the Azure provider is sourced from an environment variable that is read by a `DefaultValueFunc`. Given this, I'm going to spend a short time thinking about whether we can do better for envvar config specifically.

pgavlin:
We agreed to land this PR now. I'll follow up with some changes to improve the experience for envvar config.

PR #187, opened 2018-06-28
Only populate defaults for input properties in MakeTerraformInputs

lukehoban:
When creating terraform inputs, we would previously inject defaults for properties marked as being output-only (!Required and !Optional).

In particular, this caused AutoName transforms used in many of our TF-based providers to attempt to populate a default Name input even when the TF schema indicated Name was not an input property - leading to failures from the TF provider.

Fixes pulumi/pulumi-aws#250


pgavlin:
As discussed offline, I think we're fixing things at the wrong layer. Instead of not setting defaults for outputs, we simply shouldn't have defaults for outputs (and indeed, attempting to set one should be an error). We should be able to detect this both at `tfgen` time and in the provider itself.

lukehoban:
> I think we're fixing things at the wrong layer

Agreed - closing this one out.

PR #189, opened 2018-06-29
Include `id` property in data source outputs

### LINKED ISSUE #188: Outputs from data sources do not include `id` property
Terraform data source results will always have an ID, but we are not projecting that into a `.id` property on the output from a generated `.get` function in the terraform bridge or code generator.

For AWS and Azure, this ID is nearly always an alias for some other data projected in a separate field.  However, in the Openstack provider, this ID is frequently the only way that certain information gets returned from the data source.

Frassle:
Terraform data source results will always have an ID, this is now
projected to an attribute on the structs returned by data source get
functions.

For AWS and Azure, this ID is nearly always an alias for some other data
projected in a separate field. However, in the Openstack provider, this
ID is frequently the only way that certain information gets returned
from the data source.

As the ID field isn't exposed by terraform schema we special case it as
an attribute in the generator and provider.

Fixes #188


joeduffy:
Thank you, @Frassle -- looking now, and sorry for the delay.

@ellismg Any idea why the CI job failed to clone the scripts repo?  Is the culprit that we're using ssh-style repo URLs, rather than https-style?

```
1.18s$ git clone git@github.com:pulumi/scripts ${GOPATH}/src/github.com/pulumi/scripts
Cloning into '/home/travis/gopath/src/github.com/pulumi/scripts'...
Warning: Permanently added the RSA host key for IP address '192.30.253.112' to the list of known hosts.
Permission denied (publickey).
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.
```

pkg/tfgen/generate.go
```diff
@@ -684,6 +684,16 @@ func (g *generator) gatherDataSource(rawname string,
 		}
 	}
 
+	// Add the special "id" attribute. This isn't exposed via terraform schema, so we make one up.
+	sch := &schema.Schema{
+		Type: schema.TypeString,
+		Computed: true,
+	}
+	cust := &tfbridge.SchemaInfo{}
+	rawdoc := "id is the provider-assigned unique ID for this managed resource. It is set during deployments and may be missing (undefined) during planning phases."
```
joeduffy: I believe we can leave out the second sentence, since AFAIK data sources will always return true IDs (unlike resources that we are provisioning, where the ID might not be known during planning/previews).

ellismg:
> @ellismg Any idea why the CI job failed to clone the scripts repo? Is the culprit that we're using ssh-style repo URLs, rather than https-style?

Yes, this is exactly the problem.  I just noticed this as well with one of my PRs.  So I pushed https://github.com/pulumi/pulumi-terraform/pull/192/commits/9872759329e2747349761726b7a0186effefe88d in #192

pkg/tfgen/generate.go
Frassle: Sounds right, happy to change. Do you as a team have a preference for doing changes like that via another commit in the PR, or an amended commit?

joeduffy: Slight preference for amended to keep commit history clean, but I'm happy either way :smile:

joeduffy:
Thanks @ellismg!

ellismg:
>  Thanks @ellismg!

Should be green when @Frassle pushes another commit.

PR #192, opened 2018-07-02
Track pulumi/pulumi@master

ellismg:
Relax our constraint on pulumi/pulumi such that we track the master
branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_install:
     - gometalinter --install
 install:
     # Clone the Pulumi-wide repo so we can use its scripts.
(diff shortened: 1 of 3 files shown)


ellismg:
> Curious though - isn't this now the default, and thus we could leave out the constraint entirely? Happy to leave it there for ease of added back a version constraint though if that's the intention.

I actually thought it was unnecessary as well, at first, but `dep` prefers to lock a semantic version instead of a branch, so without the constraint, it will select the most recent tag from pulumi/pulumi.  I think we'd rather just track HEAD of master.

PR #195, opened 2018-07-03
Don't emit Removed fields in sub-structure properties

Frassle:
tfbridge/gen checked the terraform schema "Removed" property for top
level fields in resources, but it wasn't checked for fields of
sub-structures (such as "networks" on an openstack compute_instance_v2
resource).

This adds a method to do a recursive clean on resources to remove all
"Removed" attributes. The method is used by both tfbridge and tfgen when
reading from the terraform providers ResourcesMap and DataSourcesMap.

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -127,7 +127,7 @@ func (p *Provider) initResourceMaps() {
 
 		p.resources[tok] = Resource{
 			TF:       res,
-			TFSchema: p.tf.ResourcesMap[res.Name].Schema,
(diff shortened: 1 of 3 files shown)


Frassle:
Ok this isn't quite enough, the properties still show up in the webui when looking at resource details. Which I guess means they're still ending up in the state file. Will investigate some more.

Frassle:
Updated the commit. Tested with my OpenStack provider and it now correctly removes the floating_ip attribute in the network block. All other previously skipped top level attributes are also still skipped.

lukehoban:
Thanks @Frassle!

PR #197, opened 2018-07-09
Set long_description in generated setup.py

ellismg:


ellismg:
> What happens here if there is no README.rst?

When `pip clean` or the like are run, the program will crash because `README.rst` is not present.  If we want to catch the underlying exception in the `readme` function and just return an empty string, I'm happy to do that as a follow up. For now, the `make` targets that drive the build of the generated provider will ensure the file is present.

PR #202, opened 2018-07-10
Do not interpolate configuration.

### LINKED ISSUE #36: Resource parameters appear to be parsed as HIL
(originally filed as https://github.com/pulumi/pulumi/issues/376)

HIL: https://github.com/hashicorp/hil

This program cannot plan:

```typescript
import * as aws from "@pulumi/aws"

let bucket = new aws.s3.Bucket("bucket", {});

let obj = new aws.s3.BucketObject("obj", {
    bucket: bucket,
    content: "${}",
});
```

The error is pretty cryptic:
```
Planning changes:
error: An error occurred while viewing the plan: rpc error: code = Unknown desc = Error preparing urn:pulumi:pulumi-service::pulumi-service::aws:s3/bucketObject:BucketObject::obj's resource config state: parse error at 1:3: expected expression but found "}"
```

Changing `content` to `"${abc}"` works fine, and causes an S3 object to be created with the literal contents `${abc}` -- so it seems like the plan is _parsed_ as HIL but the HIL is never executed.

pgavlin:
Unlike Terraform itself, we have already applied any relevant
interpolations at the time we are creating a `ResourceConfig`. As such,
we do not need to--and indeed do not want to--pass our config values
through a `RawConfig`, which always interpolates string properties.

Fixes #36.

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -450,11 +450,10 @@ func MakeTerraformConfigFromRPC(res *PulumiResource, m *pbstruct.Struct,
 
 // MakeTerraformConfigFromInputs creates a new Terraform configuration object from a set of Terraform inputs.
 func MakeTerraformConfigFromInputs(inputs map[string]interface{}) (*terraform.ResourceConfig, error) {
-	cfg, err := config.NewRawConfig(inputs)
-	if err != nil {
-		return nil, err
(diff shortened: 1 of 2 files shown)


pgavlin:
cc @lukehoban @mmdriley

mmdriley:
lgtm

Issue #191, opened 2018-07-01
Cannot pass `Output<Archive>` when `Input<Archive>` is specified


lukehoban:
A property projected as:

```
readonly content?: pulumi.Input<pulumi.asset.Archive>;
```

will fail if a `pulumi.Output<pulumi.asset.Archive>` is passed, with this error:

```
info: panic: fatal: An assertion has failed: expected asset value for Pulumi schema info
    goroutine 185 [running]:
    github.com/pulumi/pulumi-azure/vendor/github.com/pulumi/pulumi/pkg/util/contract.failfast(0xc420019270, 0x44)
    	/Users/luke/go/src/github.com/pulumi/pulumi-azure/vendor/github.com/pulumi/pulumi/pkg/util/contract/failfast.go:23 +0xf3
    github.com/pulumi/pulumi-azure/vendor/github.com/pulumi/pulumi/pkg/util/contract.Assertf(0x280a300, 0x2a50c25, 0x2b, 0x0, 0x0, 0x0)
    	/Users/luke/go/src/github.com/pulumi/pulumi-azure/vendor/github.com/pulumi/pulumi/pkg/util/contract/assert.go:33 +0x13d
    github.com/pulumi/pulumi-azure/vendor/github.com/pulumi/pulumi-terraform/pkg/tfbridge.MakeTerraformOutput(0x272cc80, 0x2bd46b0, 0xc420482e10, 0xc4202961c0, 0xc420ecb518, 0x2a1bd00, 0x7, 0xc420482e10)
    	/Users/luke/go/src/github.com/pulumi/pulumi-azure/vendor/github.com/pulumi/pulumi-terraform/pkg/tfbridge/schema.go:339 +0xd0f
    github.com/pulumi/pulumi-azure/vendor/github.com/pulumi/pulumi-terraform/pkg/tfbridge.MakeTerraformOutputs(0xc42067d5f0, 0xc420428840, 0xc420500c60, 0xc420ecb518, 0x0, 0x0)
    	/Users/luke/go/src/github.com/pulumi/pulumi-azure/vendor/github.com/pulumi/pulumi-terraform/pkg/tfbridge/schema.go:322 +0x1bb
    github.com/pulumi/pulumi-azure/vendor/github.com/pulumi/pulumi-terraform/pkg/tfbridge.(*Provider).Check(0xc4201663c0, 0x3878c00, 0xc42067d500, 0xc42082f700, 0xc4201663c0, 0x1, 0x1)
[...]

Issue #139, opened 2018-04-10
Poor error message when passing wrong type for `BucketObjectArgs.source`


lindydonna:
I had not realized that the `source` property of `BucketObjectArgs` required a `pulumi.asset.Asset` object, and tried to use a string. I had to look up the source of the error message before I could understand what it was saying (`expected asset value for Pulumi schema info`)

## Source code
```typescript
for (let item of fs.readdirSync(path)) {
    console.log(item);
    let object = new aws.s3.BucketObject(item, { 
        bucket: siteBucket,
        source: item, // BUG: this should be 'new pulumi.asset.FileAsset'
    });
}
```

## Actual Output
```
$ pulumi preview
Previewing stack 'testing' in the Pulumi Cloud ☁️
Previewing changes:
info: index
+ pulumi:pulumi:Stack: (create)
    [urn=urn:pulumi:testing::aws-js-s3-folder::pulumi:pulumi:Stack::aws-js-s3-folder-testing]
    + aws:s3/bucket:Bucket: (create)
        [urn=urn:pulumi:testing::aws-js-s3-folder::aws:s3/bucket:Bucket::donna-bucket]
        acl         : "private"
        bucket      : "donna-bucket-51f4bc4"
        forceDestroy: false
info: panic: fatal: An assertion has failed: expected asset value for Pulumi schema info
info:
info: goroutine 77 [running]:
info: github.com/pulumi/pulumi-aws/vendor/github.com/pulumi/pulumi/pkg/util/contract.failfast(0xc420092870, 0x44)
info:   /home/travis/gopath/src/github.com/pulumi/pulumi-aws/vendor/github.com/pulumi/pulumi/pkg/util/contract/failfast.go:11 +0xf3
info: github.com/pulumi/pulumi-aws/vendor/github.com/pulumi/pulumi/pkg/util/contract.Assertf(0x32fb800, 0x386b214, 0x2b, 0x0, 0x0,0x0)
info:   /home/travis/gopath/src/github.com/pulumi/pulumi-aws/vendor/github.com/pulumi/pulumi/pkg/util/contract/assert.go:21 +0x13d
info: github.com/pulumi/pulumi-aws/vendor/github.com/pulumi/pulumi-terraform/pkg/tfbridge.MakeTerraformOutput(0x2f72260, 0xc420a44280, 0xc42021d2c0, 0xc42023d180, 0xc420719518, 0xc420a56100, 0x6, 0xc42021d2c0)
[...]

PR #205, opened 2018-07-12
Allow 'computed' assets

CyrusNajmabadi:
Fixes https://github.com/pulumi/pulumi-terraform/issues/191

Our current terraform provider assets that something typed to be an asset must actually be an asset.  However, during preview, we may have an Output<Asset|Archive>.  These will remote to the terraform provider as an 'unknown variable value'.  

This change simply updates the provider to expect that that can happen and to handle it properly (by mapping it to 'computed' value).

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -339,14 +339,13 @@ func MakeTerraformOutput(v interface{},
 
 		// we might not have the asset value if this was something computed. in that
 		// case just return an appropriate sentinel indicating that was the case.
-		contract.Assert(v != nil)
-		contract.Assert(reflect.ValueOf(v).Kind() == reflect.String)
 
-		t := v.(string)
+		t, ok := v.(string)
+		contract.Assert(ok)
 		contract.Assert(t == config.UnknownVariableValue)
 
-		elem := resource.Computed{Element: resource.NewStringProperty("")}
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/schema.go
```diff
@@ -329,11 +329,24 @@ func MakeTerraformOutputs(outs map[string]interface{},
...
+		if asset, has := assets[ps]; has {
+			// if we have the value, it better actually be an asset or an archive.
+			contract.Assert(asset.IsAsset() || asset.IsArchive())
+			return asset
+		}
+
+		// we might not have the asset value if this was something computed. in that
+		// case just return an appropriate sentinel indicating that was the case.
+		contract.Assert(v != nil)
+		contract.Assert(reflect.ValueOf(v).Kind() == reflect.String)
```
CyrusNajmabadi: two many asserts @pgavlin ?  Would it just be better to do the ```v.(string)``` cast and just have it fail if it wasn't a string?

pgavlin: I would do the following:
```
t, ok := v.(string)
contract.Assert(ok)
contract.Assert(t == config.UnknownVariableValue)
```

CyrusNajmabadi:
> Curious - does this also fix #139?

No.  However, i've talked with @pgavlin and we feel lke we can fix this as well.  i will make a followup PR.

PR #208, opened 2018-07-16
Eliminate Python keyword collisions

### LINKED ISSUE #207: Packages can conflict with Python reserved keywords
We currently can generate code using names that are Python reserved keywords.  For instance, the AWS package's `lambda` package.  Unfortunately, Python has no way to escape such names, so you need to resort to `importlib` or other dynamic means to interact with them.

We should auto-escape these names with a trailing underscore, so `lambda_`, [per PEP8](https://www.python.org/dev/peps/pep-0008/?#id46).

joeduffy:
It's possible for us to generate names that conflict with Python
keywords, leading to unusable packages and member names.  To fix this,
let's just add a trailing underscore, per PEP 8.  Fixes #207.

### DIFF
--- a/pkg/tfgen/generate_python.go
+++ b/pkg/tfgen/generate_python.go
@@ -331,7 +331,7 @@ func (g *pythonGenerator) emitRawDocComment(w *tools.GenWriter, comment, prefix 
 
 func (g *pythonGenerator) emitPlainOldType(w *tools.GenWriter, pot *plainOldType) {
 	// Produce a class definition with optional """ comment.
-	w.Writefmtln("class %s(object):", pot.name)
+	w.Writefmtln("class %s(object):", pyClassName(pot.name))
 	if pot.doc != "" {
 		g.emitDocComment(w, pot.doc, "    ")
 	}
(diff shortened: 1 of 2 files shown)


pkg/tfgen/generate_python.go
```diff
@@ -800,5 +800,59 @@ func pyName(name string) string {
 	}
 
 	components = append(components, string(currentComponent))
-	return strings.Join(components, "_")
+	result := strings.Join(components, "_")
+
+	// If the generated name clashes with a Python 2 or 3 keyword, add a trailing underscore per PEP 8:
+	// https://www.python.org/dev/peps/pep-0008/?#id46
```
pgavlin: Nit: that URL doesn't lead directly to the recommendation. This URL does: https://www.python.org/dev/peps/pep-0008/?#function-and-method-arguments

joeduffy: Ah, whoops, will fix.

joeduffy: Ah, whoops, will fix.

PR #212, opened 2018-07-19
WIP: Integration testing in Travis

ellismg:


ellismg:
I think we can land this now @lukehoban.  If you look at the status checks for this PR you'll see 3 new checks that show us the source diffs.

pkg/tfgen/generate_nodejs.go
```diff
@@ -652,7 +652,7 @@ func (g *nodeJSGenerator) emitNPMPackageMetadata(pack *pkg) error {
 			"build": "tsc",
 		},
 		DevDependencies: map[string]string{
-			"typescript": "^2.6.2",
+			"typescript": "^2.9.2",
```
ellismg: I can revert this before checking in, but I wanted some change that would cause a diff so we could see what happened.

PR #218, opened 2018-08-02
Adopt pulumi/pulumi#1684

lukehoban:
In most of the places we log within tfbridge we do not have a URN to attach the logging context to - in particular, most output from TF providers is via stdout, which we cannot associate with a specific operation.

For logging results of `Check` calls though, we can associate these with the resource that is being checked.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -385,7 +385,7 @@
     "pkg/workspace",
     "sdk/proto/go"
   ]
-  revision = "61324d554059ea96a89460d0ede30eaba4368fb3"
+  revision = "9bfea8e5c4b4b7522e00796397f72fe3f4984dd7"
 
 [[projects]]
   branch = "master"
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -66,7 +66,7 @@ type DataSource struct {
 }
 
 // NewProvider creates a new Pulumi RPC server wired up to the given host and wrapping the given Terraform provider.
-func NewProvider(host *provider.HostClient, module string, version string,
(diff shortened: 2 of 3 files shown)


lukehoban:
BTW - @pgavlin - I assume that `setLoggingContext` won't work (correctly) once we add parallelism?  I was hoping that might have been a path to attaching logging to the right URN, but I think it doesn't/won't work for the same reasons.

pgavlin:
> BTW - @pgavlin - I assume that setLoggingContext won't work (correctly) once we add parallelism? I was hoping that might have been a path to attaching logging to the right URN, but I think it doesn't/won't work for the same reasons.

Why not? Seems like it should be fine to me--this is only called once per provider instance, right?

lukehoban:
> this is only called once per provider instance, right?

No - it’s called on entry to every RPC function.

pgavlin:
> No - it’s called on entry to every RPC function.

I see it now. Yeah, that's very much not going to work. There shouldn't be any reason to do this, though, now that `Cancel` exists: we should be able to have a global logging context that uses a global `Context` that gets cancelled by `Cancel`.

lukehoban:
> I see it now. Yeah, that's very much not going to work. There shouldn't be any reason to do this, though, now that Cancel exists: we should be able to have a global logging context that uses a global Context that gets cancelled by Cancel.

I looked at this a little, but I'm not sure how exactly we'll want to structure this after parallelism and 1st class providers land.  For now, I won't change this as part of adopting pulumi/pulumi#1684, and we can clean this up as part of bringing the parallelism changes into this library.

PR #221, opened 2018-08-07
Generate provider resources for Node.

pgavlin:
These changes generate first-class provider resources for Node. Because
the base SDKs for Go and Python do not yet expose first-class providers,
these generators have not been updated.


pgavlin:
Example generated code changes from `@pulumi/aws`: https://github.com/pulumi/pulumi-aws/compare/pgavlin/providerResource

PR #230, opened 2018-08-14
Produce helper 'require' functions for config variables when generating our nodejs SDKs.

CyrusNajmabadi:
This is useful for downstream consumers of our API who want to be able to retrieve one of these values and have it blow up (with an appropriate error message) if the value is not there.  

Right now we simulate this by manually having a 'require.ts' file.  This just moves this into the generation phase so we get this for free.  Example of what this looks like is:

```ts
export let accessKey: string | undefined = __config.get("accessKey");
export function requireAccessKey(): string { return __config.require("accessKey"); }
```

As before, we still have the optional value exposed through `pulumi.config.accessKey`.  But someone can say `pulumi.config.requireAccessKey()` and if that value is not there, the user will get the appropiate error saying to set the `aws:accessKey` config value.

### DIFF
--- a/pkg/tfgen/generate_nodejs.go
+++ b/pkg/tfgen/generate_nodejs.go
@@ -300,15 +300,15 @@ func (g *nodeJSGenerator) emitConfigVariables(mod *module) (string, error) {
 }
 
 func (g *nodeJSGenerator) emitConfigVariable(w *tools.GenWriter, v *variable) {
-	var getfunc string
-	if v.optional() {
-		getfunc = "get"
-	} else {
-		getfunc = "require"
-	}
+	optionalFunc := "get"
+	requiredFunc := "require"
+
+	flagsType := tsType(v, false /*noflags*/, !v.out /*wrapInput*/)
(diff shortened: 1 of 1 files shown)


lukehoban:
I don’t think I understand the use case here. 

It should in general be rare for other packages shouldn’t need to read the config of one of these provider packages. It’s mostly config for the provider itself.  And it’s especially strange that a third party would want to treat config on some other component as required. 

I’m interested in the specific use case motivating this - but I suspect this is not common enough to motivate emitting helpers for all configuration values.

ellismg:
> I’m interested in the specific use case motivating this - but I suspect this is not common enough to motivate emitting helpers for all configuration values.

One thing I did realize looking at the diffs generated downstream is that it was not clear to me how this `vars.ts` interacts with the new first class providers stuff.  From my reading of the code the values here are always going to be what the ambient provider was configured with.  I'm not really sure how folks are expected to use this as all, but I am little worried that we may have made this even more confusing with first class providers.

CyrusNajmabadi:
Closing this out.  Luke made a good point why we likely do not want this.

PR #235, opened 2018-08-17
Ignore block docs from terraform

### LINKED ISSUE #234: tfgen picking up wrong docs from terraform
tfgen seems to not always pick the right docs from the terraform markdown. For example the following property on the openstack compute instance (https://github.com/pulumi/pulumi-openstack/blob/6fc26edd7c42e7c3d65a01cf9384148cc56466e4/sdk/nodejs/compute/instance.ts#L304). This is the name property for the instance which should of picked the following docs (from https://raw.githubusercontent.com/terraform-providers/terraform-provider-openstack/master/website/docs/r/compute_instance_v2.html.markdown):
> * `name` - (Required) A unique name for the resource.

But instead it's picking up the following same named property of a sub-property:
> * `name` - (Required unless `uuid` or `port` is provided) The human-readable
    name of the network. Changing this creates a new server.

Frassle:
Fixes #234

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -132,6 +132,9 @@ func mergeDocs(provider string, kind DocKind, targetDocs map[string]string, sour
 var argumentBulletRegexp = regexp.MustCompile(
 	"\\*\\s+`([a-zA-z0-9_]*)`\\s+(\\([a-zA-Z]*\\)\\s*)?[–-]?\\s+(\\([^\\)]*\\)\\s*)?(.*)",
 )
+var argumentBlockRegexp = regexp.MustCompile(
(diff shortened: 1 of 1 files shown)


lukehoban:
@ellismg do we have the tool working now that can produce diffs for all the tfgen-based providers with this change?  Would love to see the impact on AWS in particular.

Frassle:
aws does seem to have different doc layout to openstack and azure. I'll do some tests with it to make sure it doesn't change incorrectly.

Frassle:
Had to update to typescript dependency to pulumi 15 to get aws to build again. Here's the diff caused by just this changeset: https://gist.github.com/Frassle/5a6bb1da1758f25741c10417ccd94e7d

Picked a handful of changes to compare against the doc file and they all seem fine.

PR #236, opened 2018-08-17
Fix types and marshalling for Resource-Elem TypeMap values.

### LINKED ISSUE #231: Tfgen sometimes gets schema.TypeMaps types wrong
```javascript
"use strict";
const pulumi = require("@pulumi/pulumi");
const aws = require("@pulumi/aws");

const doc = new aws.ssm.Document("foo", {
    documentType: "Command",
    content: "somejson",
    permissions: {
        "key": {
            accountIds: "foo",
            type: "bar",
        },
    },
});
```

According to tfgen generated types, `permissions` has type `{ [key: string] { accountIds: string, type: string } }`. This isn't the right type, though:

```
    error: aws:ssm/document:Document resource 'foo' has a problem: permissions (key): '' expected type 'string', got unconvertible type 'map[string]interface {}'
```

It kinda looks like it needs to be a set instead:
```javascript
"use strict";
const pulumi = require("@pulumi/pulumi");
const aws = require("@pulumi/aws");

const doc = new aws.ssm.Document("foo", {
    documentType: "Command",
    content: "somejson",
    permissions: [
        {
            accountIds: "foo",
            type: "bar",
        },
    ],
});
```

pgavlin:
Terraform maps are odd, to say the least. In general a map value may be
either a list of string -> value maps or a string -> value map itself.
Maps that have an element type that is a *schema.Resource, however, must
be lists of string -> value maps.

These changes update the Pulumi -> TF config marshaller so that in the
latter case we appropriately wrap input maps in an array. The codegen
for Node and Go has also been updated to accurately reflect the type.

Fixes #231.

### DIFF
--- a/pkg/tfgen/generate_nodejs.go
+++ b/pkg/tfgen/generate_nodejs.go
@@ -958,7 +958,14 @@ func tsPrimitive(vt schema.ValueType, elem interface{}, eleminfo *tfbridge.Schem
 		}
 		t = fmt.Sprintf("%s[]", elemType)
 	case schema.TypeMap:
-		t = fmt.Sprintf("{[key: string]: %v}", tsElemType(elem, eleminfo, out, wrapInput))
+		// If this map has a "resource" element type, just use the generated element type. This works around a bug in
+		// TF that effectively forces this behavior.
+		elemType := tsElemType(elem, eleminfo, out, wrapInput)
+		if _, hasResourceElem := elem.(*schema.Resource); hasResourceElem {
+			t, wrapInput = elemType, false
+		} else {
+			t = fmt.Sprintf("{[key: string]: %v}", elemType)
+		}
 	default:
 		contract.Failf("Unrecognized schema type: %v", vt)
 	}
--- a/pkg/tfgen/generate_nodejs.go
+++ b/pkg/tfgen/generate_nodejs.go
@@ -962,10 +962,9 @@ func tsPrimitive(vt schema.ValueType, elem interface{}, eleminfo *tfbridge.Schem
 		// TF that effectively forces this behavior.
 		elemType := tsElemType(elem, eleminfo, out, wrapInput)
 		if _, hasResourceElem := elem.(*schema.Resource); hasResourceElem {
-			t, wrapInput = elemType, false
-		} else {
-			t = fmt.Sprintf("{[key: string]: %v}", elemType)
+			return elemType
 		}
+		t = fmt.Sprintf("{[key: string]: %v}", elemType)
 	default:
 		contract.Failf("Unrecognized schema type: %v", vt)
 	}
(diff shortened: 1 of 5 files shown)


pgavlin:
AWS diff: https://gist.github.com/pgavlin/89ba45095e98fcff4e8df8de962ede8d

lukehoban:
The AWS diff doesn’t feel right on a first look - seems we are now projecting these as scalar structs?

pgavlin:
> seems we are now projecting these as scalar structs?

That's correct. This is consistent with TF's behavior/the intended behavior of the schema. For example, this is what users are intended to write for SSM Document permissions in Terraform:
```
resource "aws_ssm_document" "foo" {
  permissions {
    "account_ids" = "all"
    "type" = "Share"
  }
}
```
The permissions field has this schema:
```
			"permissions": {
				Type:     schema.TypeMap,
				Optional: true,
				Elem: &schema.Resource{
					Schema: map[string]*schema.Schema{
						"type": {
							Type:     schema.TypeString,
							Required: true,
						},
						"account_ids": {
							Type:     schema.TypeString,
							Required: true,
						},
					},
				},
			},
```
And is used in this fashion:
```
	permission := d.Get("permissions").(map[string]interface{})

	ids := aws.StringSlice([]string{permission["account_ids"].(string)})

	if strings.Contains(permission["account_ids"].(string), ",") {
		ids = aws.StringSlice(strings.Split(permission["account_ids"].(string), ","))
	}

	permInput := &ssm.ModifyDocumentPermissionInput{
		Name:            aws.String(d.Get("name").(string)),
		PermissionType:  aws.String(permission["type"].(string)),
		AccountIdsToAdd: ids,
	}
```

The `roles` field of the Cognito identity pool role attachment type is treated similarly. This sort of usage appears to be unintentional on TF's part, FWIW--it's _much_ more common to have `TypeList`1 fields with `*schema.Resource` elements.

pkg/tfgen/generate_nodejs.go
```diff
@@ -958,7 +958,14 @@ func tsPrimitive(vt schema.ValueType, elem interface{}, eleminfo *tfbridge.Schem
 		}
 		t = fmt.Sprintf("%s[]", elemType)
 	case schema.TypeMap:
-		t = fmt.Sprintf("{[key: string]: %v}", tsElemType(elem, eleminfo, out, wrapInput))
+		// If this map has a "resource" element type, just use the generated element type. This works around a bug in
+		// TF that effectively forces this behavior.
+		elemType := tsElemType(elem, eleminfo, out, wrapInput)
+		if _, hasResourceElem := elem.(*schema.Resource); hasResourceElem {
+			t, wrapInput = elemType, false
```
lukehoban: Minor - for consistency with above, we could just `return elemType` here and drop the else (or change above to use a similar pattern as here).

pgavlin: Ah, so we could. Done.

PR #238, opened 2018-08-24
Clean up links in docs

### LINKED ISSUE #135: Help text from Terraform embeds relative links to Terraform docs
We have a several broken links in the `packages/` route on the docs site. After looking at a few most seem to be docs that have embedded HTML, such as https://docs.pulumi.com/packages/pulumi-aws/classes/_ec2_securitygroup_.securitygroup.html. Since the doc link is a relative URL, the URL https://docs.pulumi.com/packages/pulumi-aws/classes/security_group_rule.html doesn't point anywhere.

lukehoban:
Fixes #135.
Fixes pulumi/docs#494.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -200,5 +200,50 @@ func parseTFMarkdown(kind DocKind, markdown string, provider string, rawname str
 			// Ignore everything else - most commonly examples and imports with unpredictable section headers.
 		}
 	}
-	return ret
+	return cleanupDoc(ret)
+}
+
+func cleanupDoc(doc parsedDoc) parsedDoc {
+	newargs := make(map[string]string, len(doc.Arguments))
+	for k, v := range doc.Arguments {
+		newargs[k] = cleanupText(v)
+	}
+	newattrs := make(map[string]string, len(doc.Attributes))
+	for k, v := range doc.Attributes {
+		newattrs[k] = cleanupText(v)
+	}
+	return parsedDoc{
+		Description: cleanupText(doc.Description),
+		Arguments:   newargs,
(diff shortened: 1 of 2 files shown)


lukehoban:
Delta for AWS here: https://gist.github.com/lukehoban/923ce9bb3bbec798a128f68d12164584

@ellismg Do we have a working tool now to autogenerates diffs (and tests) for all TF-based providers?

pkg/tfgen/docs.go
```diff
@@ -200,5 +200,50 @@ func parseTFMarkdown(kind DocKind, markdown string, provider string, rawname str
...
+	// Find URLs and re-write local links
+	text = markdownLink.ReplaceAllStringFunc(text, func(link string) string {
+		parts := markdownLink.FindStringSubmatch(link)
+		url := parts[2]
+		if strings.HasPrefix(url, "http") {
+			// Absolute URL, return as-is
+			return link
+		} else if strings.HasPrefix(url, "/") {
+			// Relative URL to the root of the Terraform docs site, rewrite to absolute
+			return fmt.Sprintf("[%s](https://www.terraform.io%s)", parts[1], url)
```
pgavlin: Note that this will cause broken links if/when we support providers that don't have docs on terraform.io. Might be worth calling out/filing an issue.

lukehoban: Good point.  Though not exactly clear what the intent of the docs could/would be if they use `/some/path.html`-style URLs without the assumption of being hosted on `https://www.terraform.io`.  I suppose if there was some alternative hosted documentation site for Terraform providers in the future, we'd need to handle that as well.

Issue #10, opened 2017-07-20
Add a way to auto-JSON.stringify objects


joeduffy:
When we converted from hand-written resource providers to the Terraform Bridge ones, we lost the ability to pass JSON maps directly for certain properties.

For example, consider IAM policy documents.  In the old code, we would say:

    let doc = {
        Statement: [ { Sid: "", Effect: "Allow", ... } ],
    };
    let role = new iam.Role(..., { assumeRolePolicy: doc });

but now we must manually stringify the document instead:

    let role = new iam.Role(..., { assumeRolePolicy: JSON.stringify(doc) });

This is needless ceremony.  It would be easy and nice to auto-stringify certain objects.  Most likely, this would just be a `SchemaInfo` overlay setting that you could turn on per-field.

joeduffy:
I'll take a whack at this one -- a very valued community member says this will improve quality of life, which I am obviously very interested in doing 😄

PR #239, opened 2018-08-29
Add the ability to auto-transform values

joeduffy:
This change adds the ability to auto-transform values on the fly,
in between unmarshaling from an RPC call, and feeding said values
into Terraform. This allows us to achieve pulumi/pulumi-terraform#10,
which is to automatically JSON.stringify values, letting us put strong
typing annotations on some of our provider interfaces (like IAM docs).

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -64,7 +64,7 @@ type SchemaInfo struct {
 	Name        string                 // a name to override the default; "" uses the default.
 	Type        tokens.Type            // a type to override the default; "" uses the default.
 	AltTypes    []tokens.Type          // alternative types that can be used instead of the override.
-	Transform   Transformer            // an optional transformation func, applied before passing to TF.
+	Transform   Transformer            // an optional idemponent transformation, applied before passing to TF.
 	Elem        *SchemaInfo            // a schema override for elements for arrays, maps, and sets.
 	Fields      map[string]*SchemaInfo // a map of custom field names; if a type is missing, the default is used.
 	Asset       *AssetTranslation      // a map of asset translation information, if this is an asset.
@@ -73,7 +73,12 @@ type SchemaInfo struct {
 	MaxItemsOne *bool                  // to override whether this property should project as a scalar or array.
 }
 
-// Transformer is given the option to transform a value in situ before it is processed by the bridge.
+// Transformer is given the option to transform a value in situ before it is processed by the bridge. This
+// transformation must be deterministic and idempotent, and any value produced by this transformation must
+// be a legal alternative input value. A good example is a resource that accepts either a string or
+// JSON-stringable map; a resource provider may opt to store the raw string, but let users pass in maps as
+// a convenience mechanism, and have the transformer stringify them on the fly. This is safe to do because
+// the raw string is still accepted as a possible input value.
 type Transformer func(resource.PropertyValue) (resource.PropertyValue, error)
 
 // DocInfo contains optional overrids for finding and mapping TD docs.


joeduffy:
We probably could, but in the cases we will use this, it'll be with a union type; e.g., accept `string | iam.PolicyDocument` as input. We already have this issue today where the type on the output side may only be one of these things (think ARNs versus strongly typed resources).

joeduffy:
@pgavlin We had talked about this a bunch in the past, so I'm going to go ahead and merge so I can adopt this downstream. Please shout if my answer didn't qualm your concerns.

lukehoban:
> We already have this issue today where the type on the output side may only be one of these things (think ARNs versus strongly typed resources).

But on the wire (and in the checkpoint), these were always the same, right?  We kind of relied on the fact that the we could only use these strongly typed AltType wrappers on things where the stronger type didn't actually change how the value was represented (replacing a place where an ARN might have been expected, but we allowed passing a strongly typed resource which would get rewritten as an ARN in the Pulumi object model representations).  

I think the concern is that we are newly introducing an ability to have these things truly diverge.  It's not immediately clear to me that this will not cause deeper problems given current resource model. 

Have we verified that this doesn't lead to these properties constantly reporting diffs even when there are no changes?  @pgavlin Other things we should verify here?

joeduffy:
Yes, I have verified that it works as expected

pgavlin:
One of the things I'm currently working through is whether or not our "Pulumi inputs -> provider inputs -> provider state" model is workable in a world with engine diffing, read, refresh, and importing. The most critical challenge that this model introduces is that it is not necessarily possible to generate the provider/Pulumi inputs for a resource given its current provider state. As such, I'm reluctant to take changes that affect this scenario.

All that said, this PR may be okay--I just need to think through its implications a bit more closely.

lukehoban:
> it is not necessarily possible to generate the provider/Pulumi inputs for a resource given its current provider state. 

I think the good news for this change is that it doesn't make this situation worse.  You can still generate valid inputs from the provider state.  In fact, we should perhaps add a comment on the `Transform` function that it is expected that it is idempotent - anything it can return should be a valid input (on which the transform is the identity function) as well.  The `TransformJSONDocument` example already ensures this - but I could see it being tempting to use this in ways that don't.

pgavlin:
> I could see it being tempting to use this in ways that don't.

Yes, definitely. I'm half-tempted to push this kind of transformation up into the SDKs, but that does seem like it's probably unnecessarily complex.

lukehoban:
As long as we document the requirements for correctness, I think it's fine to provide powerful primitives at the `pulumi-terraform` layer.  And most likely net result of doing it once at this layer will be more consistency.

pkg/tfbridge/info.go
```diff
@@ -72,6 +73,9 @@ type SchemaInfo struct {
 	MaxItemsOne *bool                  // to override whether this property should project as a scalar or array.
 }
 
+// Transformer is given the option to transform a value in situ before it is processed by the bridge.
```
lukehoban: Perhaps add a comment that it is expected that the Transformer be idempotent (the identity function on any value that is a non-error output)?

joeduffy: Great idea. I added a comment and pushed:

```
// Transformer is given the option to transform a value in situ before it is processed by the bridge. This
// transformation must be deterministic and idempotent, and any value produced by this transformation must
// be a legal alternative input value. A good example is a resource that accepts either a string or
// JSON-stringable map; a resource provider may opt to store the raw string, but let users pass in maps as
// a convenience mechanism, and have the transformer stringify them on the fly. This is safe to do because
// the raw string is still accepted as a possible input value.
```

PR #240, opened 2018-08-30
Project default values from literals or envvars.

pgavlin:
It is common for provider properties in Terraform to have defaults that
are pulled from environment variables or a literal value. These changes
add support for this scenario by adding an additional property to the
schema overlay containing envvars for default values. This property is
then used in concert with the existing literal default to generate
appropriate code for each language.

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -95,6 +95,10 @@ func MakeTerraformInputs(res *PulumiResource, olds, news resource.PropertyMap,
 					}
 					result[name] = v
 					glog.V(9).Infof("Created Terraform input: %v = %v (old default)", key, old)
+				} else if envVars := info.Default.EnvVars; len(envVars) != 0 {
+					result[name] = schema.MultiEnvDefaultFunc(envVars, info.Default.Value)
+					glog.V(9).Infof("Created Terraform input: %v = %v (defualt from env vars)", name, result[name])
+
 				} else if info.Default.Value != nil {
 					result[name] = info.Default.Value
 					glog.V(9).Infof("Created Terraform input: %v = %v (default)", name, result[name])
@@ -103,7 +107,6 @@ func MakeTerraformInputs(res *PulumiResource, olds, news resource.PropertyMap,
 					if err != nil {
 						return nil, err
 					}
-
 					result[name] = v
 					glog.V(9).Infof("Created Terraform input: %v = %v (default from fnc)", name, result[name])
 				}
@@ -603,6 +606,7 @@ func MakeTerraformAttributesFromInputs(inputs map[string]interface{},
 
 		flattenValue(result, k, f.Value)
 	}
+
 	return result, nil
 }
 
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -97,7 +97,7 @@ func MakeTerraformInputs(res *PulumiResource, olds, news resource.PropertyMap,
 					glog.V(9).Infof("Created Terraform input: %v = %v (old default)", key, old)
 				} else if envVars := info.Default.EnvVars; len(envVars) != 0 {
 					result[name] = schema.MultiEnvDefaultFunc(envVars, info.Default.Value)
-					glog.V(9).Infof("Created Terraform input: %v = %v (defualt from env vars)", name, result[name])
+					glog.V(9).Infof("Created Terraform input: %v = %v (default from env vars)", name, result[name])
 
 				} else if info.Default.Value != nil {
 					result[name] = info.Default.Value
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -208,7 +208,7 @@ type variable struct {
 	name   string
 	out    bool
 	opt    bool
-	config bool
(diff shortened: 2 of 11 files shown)


pgavlin:
Example AWS diff: https://github.com/pulumi/pulumi-aws/compare/pgavlin/envDefaults

pgavlin:
Interesting files to look at in that diff are `sdk/nodejs/provider.ts` and `sdk/nodejs/config/vars.ts`. We _might_ need to change things around a bit s.t. we can have `require`-like behavior if no default is found in an envvar.

pkg/tfbridge/schema.go
```diff
@@ -95,6 +95,10 @@ func MakeTerraformInputs(res *PulumiResource, olds, news resource.PropertyMap,
 					}
 					result[name] = v
 					glog.V(9).Infof("Created Terraform input: %v = %v (old default)", key, old)
+				} else if envVars := info.Default.EnvVars; len(envVars) != 0 {
+					result[name] = schema.MultiEnvDefaultFunc(envVars, info.Default.Value)
+					glog.V(9).Infof("Created Terraform input: %v = %v (defualt from env vars)", name, result[name])
+
```
lukehoban: Nit: extra empty line

pkg/tfgen/generate.go
```diff
@@ -203,6 +208,7 @@ type variable struct {
 	name   string
 	out    bool
 	opt    bool
+	config bool
```
lukehoban: None of the rest have comments - but this would have been a lot easier to follow if we commented here what `config = true` means.

pkg/tfgen/generate_go_utilities.go
```diff
@@ -0,0 +1,45 @@
+package tfgen
+
+const goUtilitiesFile = `
```
lukehoban: I guess we don't have an "overlays" mechanism for our Go code generator?  Surprised we can't just read this file as a `.go` file off disk and place it in the desired location.

pkg/tfgen/generate_nodejs_utilities.go
```diff
@@ -0,0 +1,37 @@
+package tfgen
+
+const tsUtilitiesFile = `
```
lukehoban: I was surprised we aren't using the overlay system to allow this to be authored in normal `.ts` files.  Is that because overlays can only be provided via the end user of tfgen?

pkg/tfgen/generate_nodejs_utilities.go
```diff
@@ -0,0 +1,37 @@
...
+            return value;
+        }
+    }
+    return undefined;
+}
+
+export function getEnvBoolean(...vars: string[]): boolean | undefined {
+    const s = getEnv(...vars);
+    if (s !== undefined) {
+        if (["1", "t", "T", "true", "TRUE", "True"].find(v => v === s) !== undefined) {
```
lukehoban: I guess this list comes from TF (or from Go library behaviour somewhere?)?  Can we cross reference with the place this is defined?

pkg/tfgen/generate_go.go
```diff
@@ -253,11 +266,41 @@ func (g *goGenerator) emitConfigAccessor(w *tools.GenWriter, v *variable) {
...
+		defaultValue = g.goDefaultValue(v)
+	}
+
 	w.Writefmtln("func Get%s(ctx *pulumi.Context) %s {", upperFirst(v.name), gettype)
-	w.Writefmtln("\treturn config.%s(ctx, \"%s:%s\")", getfunc, g.pkg, v.name)
+
+	configKey := fmt.Sprintf("\"%s:%s\"", g.pkg, v.name)
+	if defaultValue != "" {
+		w.Writefmtln("\tif v := config.Get(ctx, %s); v != \"\" {", configKey)
+		w.Writefmtln("\t\treturn config.%s(ctx, %s)", getfunc, configKey)
```
lukehoban: In the generated AWS diff, it was surprising that this leads to duplicated `config.Get(ctx, "aws:region")`.  Not a big deal though.

pkg/tfgen/generate_go_utilities.go
```diff
@@ -0,0 +1,45 @@
...
+	for _, v := range vars {
+		if value := os.Getenv(v); value != "" {
+			if parser != nil {
+				return parser(value)
+			}
+			return value
+		}
+	}
+	return def
+}`
```
lukehoban: Trailing newline?

pkg/tfgen/generate_nodejs_utilities.go
pgavlin: This comes from Go library behavior. I will add a reference to the relevant function.

pkg/tfgen/generate_nodejs_utilities.go
pgavlin: > Is that because overlays can only be provided via the end user of tfgen?

Yes.

pkg/tfgen/generate.go
pgavlin: True--I will add a comment.

pgavlin:
> Yes - I think we do still need to have the same early failure if no value is available even after looking for environment-based defaults.

Okay. That is going to be a bit tricky, but we should be able to make it work.

PR #242, opened 2018-08-31
Adopt change from pulumi/pulumi#1853

lukehoban:


lukehoban:
Merging this now to unblock - if we decide to clean up the API here in the future we can potentially undo these changes.

Issue #226, opened 2018-08-09
'Read'ing an azure.storage.Account crashes in terraform during migration.


CyrusNajmabadi:
I have the following TS code in my app (inside @pulumi/cloud-azure to be specific):

```ts
            const result = await azure.storage.getAccount({
                name: storageAccountName,
                resourceGroupName: resourceGroupName,
            });

            console.log("getAccount result: " + JSON.stringify(result, null, 2));

            const account = azure.storage.Account.get(name, result.id);
```

The call to .getAccount succeeds and returns useful data.  However, the call to `Account.get` ends up crashing with the following stack:

<details>

```
info: panic: runtime error: index out of range
    goroutine 80 [running]:
    github.com/pulumi/pulumi-azure/vendor/github.com/terraform-providers/terraform-provider-azurerm/azurerm.migrateStorageAccountStateV0toV1(0xc4207aa3c0, 0x1, 0x1, 0x0)
    	/home/cyrusn/go/src/github.com/pulumi/pulumi-azure/vendor/github.com/terraform-providers/terraform-provider-azurerm/azurerm/resource_arm_storage_account_migration.go:37 +0x338
    github.com/pulumi/pulumi-azure/vendor/github.com/terraform-providers/terraform-provider-azurerm/azurerm.resourceStorageAccountMigrateState(0x0, 0xc4207aa3c0, 0x1f9f940, 0xc420bac000, 0x300a7a0, 0x1ed1f00, 0x40d872)
    	/home/cyrusn/go/src/github.com/pulumi/pulumi-azure/vendor/github.com/terraform-providers/terraform-provider-azurerm/azurerm/resource_arm_storage_account_migration.go:17 +0x1d1
    github.com/pulumi/pulumi-azure/vendor/github.com/hashicorp/terraform/helper/schema.(*Resource).Refresh(0xc4205f7f80, 0xc4207aa3c0, 0x1f9f940, 0xc420bac000, 0xc4203d2b30, 0x1, 0x1ed1fa0)
    	/home/cyrusn/go/src/github.com/pulumi/pulumi-azure/vendor/github.com/hashicorp/terraform/helper/schema/resource.go:340 +0x293
    github.com/pulumi/pulumi-azure/vendor/github.com/hashicorp/terraform/helper/schema.(*Provider).Refresh(0xc4205ffb90, 0xc420633600, 0xc4207aa3c0, 0xc4203ee2a0, 0x0, 0xc4206e4360)
[...]

PR #244, opened 2018-09-05
Save TF resource metadata in state.

pgavlin:
These changes add support for storing a Terraform resource's metadata
alongside its attributes in its Pulumi state. The metadata is stored as
a Pulumi property map under the key `__meta`, which should not be used
as a property name by a Terraform resource. Notably, this metadata
includes the `schema_version` property, which is used by the TF schema
code to detect the need for schema migration.

If a resource has no metadata but its schema's `SchemaVersion` is
non-zero, we inject that version into its metadata before interacting
with the TF provider. This ensures that TF always sees a valid schema
version when one is required.

This is necessary to address #226.

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -313,6 +313,9 @@ func MakeTerraformInput(res *PulumiResource, name string,
 	}
 }
 
+// metaKey is the key in a TF bridge result that is used to store a resource's meta-attributes.
+const metaKey = "__meta"
+
 // MakeTerraformResult expands a Terraform state into an expanded Pulumi resource property map.  This respects
 // the property maps so that results end up with their correct Pulumi names when shipping back to the engine.
 func MakeTerraformResult(state *terraform.InstanceState,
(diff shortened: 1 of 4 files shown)


pkg/tfbridge/schema.go
```diff
@@ -491,25 +500,45 @@ func MakeTerraformConfigFromInputs(inputs map[string]interface{}) (*terraform.Re
...
+	} else if res.SchemaVersion > 0 {
+		// If there was no metadata in the input and this resource has a non-zero schema version, return a meta bag
+		// with the current schema version. This helps avoid migration issues.
+		meta = map[string]interface{}{"schema_version": strconv.Itoa(res.SchemaVersion)}
+	}
+
 	// Turn the resource properties into a map.  For the most part, this is a straight Mappable, but we use MapReplace
 	// because we use float64s and Terraform uses ints, to represent numbers.
-	inputs, err := MakeTerraformInputs(res, nil, m, tfs, ps, nil, defaults, false)
+	inputs, err := MakeTerraformInputs(nil, nil, m, tfs, ps, nil, defaults, false)
```
lukehoban: Why is this now passing `nil`?

pgavlin: The input this used to pass-`res`--was always `nil` prior to this change, and is now of a different type.

pgavlin:
> Will refresh for the first time after picking up this change show every resource having a change?

Very possibly, yes.

> If the schema version increases for a resource, will we observe a diff and/or force a TF migration?

Yes. I believe that this is the correct behavior, and is consistent with TF.

> What is typically in this meta bag?

AFAICT the most common piece of data in this bag is the schema version.

pgavlin:
>> Will refresh for the first time after picking up this change show every resource having a change?
> Very possibly, yes.

(and I should point out that we have no way to avoid this without special-casing things in the CLI or changing the resource model)

PR #247, opened 2018-09-06
Support timeouts in TF metadata.

### LINKED ISSUE #151: Timeouts configured in Terraform resources don't take effect
e.g. an attempt to create an `aws.rds.ClusterInstance` failed after 20 minutes:
```
    error: Plan apply failed: creating urn:pulumi:mattdr::pulumi-service::aws:rds/cluster:Cluster::databaseCluster2: timeout while waiting for state to become 'available' (last state: 'creating', timeout: 20m0s)
```

this should have used a 120 minute timeout:

https://github.com/pulumi/terraform-provider-aws/blob/bccccac6a6b780b7dc1fda31d5e6d323370d4f86/aws/resource_aws_rds_cluster.go#L353
https://github.com/pulumi/terraform-provider-aws/blob/bccccac6a6b780b7dc1fda31d5e6d323370d4f86/aws/resource_aws_rds_cluster.go#L30-L32

but instead it seems to have ended up using a "system default" timeout:
https://github.com/hashicorp/terraform/blob/0c39b69d50ad6d54c4fe61d310cd45183dbf9b8c/helper/schema/resource_data.go#L369-L370

pgavlin:
Now that we are recording TF metadata, we need to handle custom
timeouts. These are encoded as int64 values in the metadata bag.

Fixes #151.

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -401,8 +401,10 @@ func MakeTerraformOutput(v interface{},
 	switch val.Kind() {
 	case reflect.Bool:
 		return resource.NewBoolProperty(val.Bool())
-	case reflect.Int:
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
 		return resource.NewNumberProperty(float64(val.Int()))
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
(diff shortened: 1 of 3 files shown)


pgavlin:
This fixes the root cause of the failures in https://github.com/pulumi/pulumi-aws/pull/315 et. al.

pgavlin:
I'm going to pre-emptively merge this to unblock pushing `pulumi-terraform` upstream. I'll follow up on any feedback after the fact.

PR #248, opened 2018-09-06
Encode TF metadata as JSON.

pgavlin:
This is consistent with TF's own encoding, and saves us from having to
special-case number encoding and decoding for TF metadata.


pgavlin:
Going to go ahead and merge; I'll address feedback after the fact.

PR #249, opened 2018-09-06
Fix envvar default reads in tfbridge.

pgavlin:
- MultiEnvDefaultFunc returns a function that must be invoked to produce
  a default value. We were not invoking the function.
- The return value of the default-producing function may be a string
  that needs to be converted to an appropriate primitive type.


pgavlin:
@lukehoban PTAL. I will add test coverage momentarily.

PR #250, opened 2018-09-06
Marshal/unmarshal complex config as JSON.

### LINKED ISSUE #48: Can't set provider variables to anything but string
Trying to set this variable:
https://github.com/terraform-providers/terraform-provider-aws/blob/ad3cd7911763d4b0b78977e6fcdfc6566350c6e9/aws/provider.go#L44

as described [here](https://www.terraform.io/docs/providers/aws/#assume-role), using this command line:
`pulumi config text aws:config:assumeRole '[{"roleArn": "arn:aws:iam::058607598222:role/OrganizationAdministrator"}]'`

(See #37 for why this is an array.)

This configuration doesn't seem to have any effect.

```
$ pulumi preview --logtostderr -v=3
...
I1031 14:35:31.890738   51605 eventsink.go:64] No assume_role block read from configuration
...
```

(That log message is [here](https://github.com/terraform-providers/terraform-provider-aws/blob/ad3cd7911763d4b0b78977e6fcdfc6566350c6e9/aws/provider.go#L654).)

One guess is this is related to this comment in `MakeTerraformInput `:
https://github.com/pulumi/pulumi-terraform/blob/6f690f52a698c9a7638cea9c315760fd691f9c91/pkg/tfbridge/schema.go#L85

since [`assume_role` is a `schema.Set`](https://github.com/terraform-providers/terraform-provider-aws/blob/ad3cd7911763d4b0b78977e6fcdfc6566350c6e9/aws/provider.go#L694).

pgavlin:
Terraform providers accept complex (i.e. non-string) config. As it
stands, we require that all provider configuration values are strings,
which causes an unfortunate inability to use these complex configuration
values. These changes enable the use of complex configuration by
serializing and deserializing complex config values using JSON.

Note that this creates a compat issue for future changes to enable
actual rich configuration in the RPC interface; such a future change
will need to accept rich values _or_ strings for top-level properties.

Fixes #48.

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -15,6 +15,7 @@
 package tfbridge
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 	"strings"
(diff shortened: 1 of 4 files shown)


pgavlin:
When using first-class providers, this looks like:
```typescript
import * as pulumi from "@pulumi/pulumi";
import * as aws from "@pulumi/aws";

const testing = new aws.Provider("testing", {
    region: aws.config.region,
    assumeRole: {
        roleArn: "<redacted>",
    },
});

const staging = new aws.Provider("staging", {
    region: aws.config.region,
    assumeRole: {
        roleArn: "<redacted>",
    },
});

new aws.s3.Bucket("testing-bucket", {}, { provider: testing });
new aws.s3.Bucket("staging-bucket", {}, { provider: staging });
```

When using the config system, this is a bit uglier, but it works:
```bash
$ pulumi config set aws:assumeRole "{ \"role_arn\": \"<redacted>\" }"
```

pgavlin:
(this will fix https://github.com/pulumi/pulumi-aws/issues/252 when rolled out to `pulumi-aws`)

PR #251, opened 2018-09-08
Two small fixes.

pgavlin:
- Make complex config serialization work with outputs.
- Appease `vetshadow`.

### DIFF
--- a/pkg/tfgen/generate_nodejs_utilities.go
+++ b/pkg/tfgen/generate_nodejs_utilities.go
@@ -1,6 +1,8 @@
 package tfgen
 
 const tsUtilitiesFile = `
+import * as pulumi from "@pulumi/pulumi";
+
 export function getEnv(...vars: string[]): string | undefined {
     for (const v of vars) {
         const value = process.env[v];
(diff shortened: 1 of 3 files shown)


pkg/tfgen/generate_nodejs_utilities.go
```diff
@@ -47,4 +49,30 @@ export function requireWithDefault<T>(req: () => T, def: T | undefined): T {
...
+    if (val === undefined || typeof val !== 'object') {
+        return val;
+    }
+
+    // Recurse on outputs, promises, arrays, and objects.
+    if (pulumi.Output.isInstance(val)) {
+        return val.apply(unwrap);
+    }
+    if (val instanceof Promise || val instanceof Array) {
+        return pulumi.output(val).apply(unwrap);
```
CyrusNajmabadi: i'm surprised this works for Array. Something doesn't feel right htere.  I would think we'd first need to call .map(unwrap) on the array elements, then do a pulumi.all on the result of that.

pkg/tfgen/generate_nodejs_utilities.go
```diff
@@ -47,4 +49,30 @@ export function requireWithDefault<T>(req: () => T, def: T | undefined): T {
     }
     return def;
 }
+
+export function unwrap(val: pulumi.Input<any>): pulumi.Input<any> {
+    // Bottom out at primitives.
+    if (val === undefined || typeof val !== 'object') {
+        return val;
+    }
```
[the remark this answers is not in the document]
CyrusNajmabadi: it's very idiomatic to do things like:

```ts
if (val === null) {
    return null;
}

switch (typeof val) {
   case "string": case "number": case "boolean": case "undefined": return val;
}
```

Then, at this point, you have the interesting objects left.

pkg/tfgen/generate_nodejs_utilities.go
pgavlin: It doesn't--we do need `map`. This was a bad edit (side note: I wish we could use an actual TS file here).

PR #254, opened 2018-09-14
Fix the token for provider resources.

pgavlin:
The token must be the same as the Pulumi package name, not the TF
provider name.


pgavlin:
> Are we just lucky this worked for the providers we have been using?

Yep.

PR #258, opened 2018-09-21
Handle computed values in the JSON transformer.

pgavlin:
If the value presented to the JSON transformer is or contains any
computed properties, return a computed string property.

### DIFF
--- a/pkg/tfbridge/transforms.go
+++ b/pkg/tfbridge/transforms.go
@@ -16,7 +16,6 @@ package tfbridge
 
 import (
 	"encoding/json"
-	"reflect"
 
 	"github.com/pkg/errors"
 	"github.com/pulumi/pulumi/pkg/resource"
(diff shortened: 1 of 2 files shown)


pgavlin:
I wavered over whether or not to run transforms at all for properties that contain unknown values, and decided that it does seem useful: this could be used to e.g. deal with properties that must always have a value. Happy to flip on this if the other option seems better--not running transforms at all is more consistent with the SDK, after all (and probably less of a minefield).

pgavlin:
Travis appears to somehow have tested the same commit twice. Closing and reopening in the hopes of triggering a proper build.

pkg/tfbridge/transforms.go
```diff
@@ -25,6 +24,11 @@ import (
 // TransformJSONDocument permits either a string, which is presumed to represent an already-stringified JSON document,
 // or a map, which will be transformed into its JSON representation.
 func TransformJSONDocument(v resource.PropertyValue) (resource.PropertyValue, error) {
+	// We can't marshal properties that contain unknowns. Turn these into an unknown value instead.
+	if v.ContainsUnknowns() {
+		return resource.MakeComputed(resource.NewStringProperty("")), nil
```
lukehoban: I'm not sure I follow the implications of returning `""` here - is that the same as what the value would have looked like prior to us adding this `TransformJSONDocument`?

pgavlin: We're not returning `""`; we're returning `computed<>`. The `""` is just a formality, and will never be displayed.

PR #259, opened 2018-09-22
Enable "destination overlays".

pgavlin:
A destination overlay is an overlay that is expected to be present in
the destination directory a priori. Like a normal overlay, it is added
to its module's index.ts.

### DIFF
--- a/pkg/tfgen/generate_nodejs.go
+++ b/pkg/tfgen/generate_nodejs.go
@@ -17,6 +17,7 @@ package tfgen
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"path/filepath"
 	"reflect"
 	"sort"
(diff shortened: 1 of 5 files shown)


pkg/tfgen/generate_nodejs.go
```diff
@@ -675,8 +676,14 @@ func (g *nodeJSGenerator) emitOverlay(mod *module, overlay *overlayFile) (string
...
-		return "", err
+	if overlay.Copy() {
+		if err := copyFile(overlay.src, dst); err != nil {
+			return "", err
+		}
+	} else {
+		if _, err := os.Stat(dst); err != nil {
+			return "", err
+		}
 	}
```
CyrusNajmabadi: my only mild feeback is that i don't like that this is effectivly identical to the code in generate_go, and it would be nice to share hte logic.

pkg/tfgen/generate_python.go
```diff
@@ -577,8 +578,14 @@ func (g *pythonGenerator) emitOverlay(mod *module, overlay *overlayFile) (string
...
 	dst := filepath.Join(dir, overlay.name)
-	if err := copyFile(overlay.src, dst); err != nil {
-		return "", err
+	if overlay.Copy() {
+		if err := copyFile(overlay.src, dst); err != nil {
+			return "", err
+		}
+	} else {
+		if _, err := os.Stat(dst); err != nil {
+			return "", err
```
CyrusNajmabadi: here too.

PR #264, opened 2018-10-11
remove ref to null README.rst, use VERSION instead

slikk66:


ellismg:
@slikk66 Not 100% sure on if this is the right fix.  Can you see my comment on pulumi/pulumi-azure#132 and help me understand how you are running into this state?

jen20:
Thanks @slikk66 - a fix for this is included in #500. Thanks for opening a PR!

PR #268, opened 2018-10-29
[WIP] Add node.js module containing Terraform interpolation functions

### LINKED ISSUE #267: Create a package for built-in functions
We decided to create a package, `@pulumi/terraform`, which contains all of the [Terraform built-in functions](https://www.terraform.io/docs/configuration/interpolation.html#built-in-functions) in raw, unopinionated form (same names, same semantics). This will help significantly with conversion tooling and efforts (for instance, `tf2pulumi` can use it).

jen20:
This PR is work in progress to fix #267.

The interpolation functions in Terraform are as follows (checked off when part of this PR, or with notes beside them):

- [x] `abs`
- [x] `basename`
- [x] `base64decode`
- [x] `base64encode`
- [ ] `base64gzip` - implemented but `gzip` options appear different so test cases do not pass 1-1
- [x] `base64sha256`
- [x] `base64sha512`
- [x] `bcrypt`
- [x] `ceil`
- [x] `chomp`
- [x] `chunklist`
- [ ] `cidrhost`
- [ ] `cidrnetmask`
- [ ] `cidrsubnet`
- [x] `coalesce`
- [x] `coalescelist`
- [x] `compact`
- [x] `concat`
- [x] `contains`
- [x] `dirname`
- [x] `distinct`
- [x] `element`
- [x] `file`
- [ ] `flatten`
- [x] `floor`
- [x] `format`
- [x] `formatlist`
- [x] `indent`
- [x] `index`
- [x] `join`
- [x] `jsonencode`
- [x] `keys`
- [x] `length`
- [x] `list`
- [x] `log`
- [x] `lower`
- [x] `map`
- [x] `max`
- [ ] `matchkeys`
- [x] `md5`
- [x] `merge`
- [x] `min`
- [x] `pathexpand`
- [x] `pow`
- [x] `uuid`
- [x] `replace`
- [ ] `rsadecrypt`
- [x] `sha1`
- [x] `sha256`
- [x] `sha512`
- [x] `signum`
- [x] `slice`
- [x] `sort`
- [x] `split`
- [x] `substr`
- [x] `timestamp`
- [ ] `timeadd`
- [x] `title`
- [x] `transpose`
- [x] `trimspace`
- [x] `upper`
- [x] `values`
- [x] `urlencode`
- [x] `zipmap`

There is a minimal set of runtime dependencies (and several more development dependencies).

Still do do:

- [ ] Implement remaining functions
- [ ] Ensure appropriate documentation comments are included
- [ ] Ensure the package build is hooked into the build system
- [ ] Ensure all tests pass and are hooked into the build system

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ TESTPARALLELISM = 10
 build::
 	go build ${PROJECT}/pkg/tfgen
 	go build ${PROJECT}/pkg/tfbridge
+	cd sdk/nodejs && npm install && npm run build
 
 lint::
 	golangci-lint run
@@ -18,6 +19,7 @@ test_fast::
 
 test_all::
 	$(GO_TEST) ${GOPKGS}
+	cd sdk/nodejs && npm run test
 
 .PHONY: publish_packages
 	$(call STEP_MESSAGE)
--- a/Makefile
+++ b/Makefile
@@ -22,12 +22,13 @@ test_all::
 	cd sdk/nodejs && npm run test
 
 .PHONY: publish_packages
+publish_packages::
 	$(call STEP_MESSAGE)
 	./scripts/publish_packages.sh
 
 # The travis_* targets are entrypoints for CI.
 .PHONY: travis_cron travis_push travis_pull_request travis_api
 travis_cron: all
-travis_push: all
+travis_push: all publish_packages
 travis_pull_request: all
 travis_api: all
--- a/Makefile
+++ b/Makefile
@@ -14,9 +14,14 @@ lint::
 
 test_fast::
 	$(GO_TEST_FAST) ${GOPKGS}
+	cd 
 
 test_all::
 	$(GO_TEST) ${GOPKGS}
+
+.PHONY: publish_packages
+	$(call STEP_MESSAGE)
+	./scripts/publish_packages.sh
 
 # The travis_* targets are entrypoints for CI.
 .PHONY: travis_cron travis_push travis_pull_request travis_api
--- a/scripts/publish_packages.sh
+++ b/scripts/publish_packages.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+# publish.sh builds and publishes a release.
+set -o nounset -o errexit -o pipefail
+ROOT=$(dirname $0)/..
+
+echo "Publishing NPM packages to NPMjs.com:"
+
+# For each package, first create the package.json to publish.  This must be different than the one we use for
+# development and testing the SDK, since we use symlinking for those workflows.  Namely, we must promote the SDK
+# dependencies from peerDependencies that are resolved via those links, to real installable dependencies.
+publish() {
+    node $(dirname $0)/promote.js ${@:2} < \
+        ${ROOT}/sdk/nodejs/$1/bin/package.json > \
+        ${ROOT}/sdk/nodejs/$1/bin/package.json.publish
+    pushd ${ROOT}/sdk/nodejs/$1/bin
+    mv package.json package.json.dev
+    mv package.json.publish package.json
+
+    NPM_TAG="dev"
+
+    # We use the "features/" prefix on branches that we want to build and publish for doing
+    # cross repo work. But in this case, we don't want to publish over "dev", since the bits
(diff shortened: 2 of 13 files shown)


sdk/nodejs/index.ts
```diff
@@ -0,0 +1,489 @@
...
+    return Math.abs(input);
+}
+
+/**
+ * basename returns the last element of a path.
+ *
+ * @param path the full path
+ */
+export function basename(path: string): string {
+    return Path.win32.basename(path);
```
lukehoban: Why does this use win32 basename?

sdk/nodejs/index.ts
```diff
@@ -0,0 +1,489 @@
...
+    return decoded;
+}
+
+/**
+ * base64encode returns a base64-encoded version of the given string.
+ *
+ * @param unencoded the string to encode
+ */
+export function base64encode(unencoded: string): string {
+    return new Buffer(unencoded).toString("base64");
```
lukehoban: I believe `new Buffer` is deprecated in Node 10.  We should probably use `Buffer.from`.

https://nodesource.com/blog/understanding-the-buffer-deprecation-in-node-js-10/

sdk/nodejs/index.ts
```diff
@@ -0,0 +1,489 @@
...
+export function dirname(path: string): string {
+    return Path.win32.dirname(path);
+}
+
+export function distinct(input: string[]): string[] {
+    return input.filter((elem, idx) => input.indexOf(elem) === idx);
+}
+
+export function element(inputList: string[], elementIndex: number) {
+    if (list.length === 0) {
```
lukehoban: Should this be `inputList`?  Not actually sure how this is legal TS, I guess there is a global `list` somewhere in builtins?

sdk/nodejs/index.ts
```diff
@@ -0,0 +1,489 @@
...
+    } else {
+        return 1;
+    }
+}
+
+export function upper(input: string): string {
+    return input.toUpperCase();
+}
+
+export function transpose(inputMap: { [k: string]: Array<string> }): { [k: string]: string[] } {
```
lukehoban: Any reason for `Array<string>` vs. `string[]` in the signature here (they should be the same, and the latter is also used later in the signature.

sdk/nodejs/tests/tests.ts
```diff
@@ -0,0 +1,550 @@
...
+const validator = require("validator");
+
+describe("abs", () => {
+    it("works", () => {
+        assert.equal(sut.abs(1), 1);
+        assert.equal(sut.abs(-1), 1);
+        assert.equal(sut.abs(-3.14), 3.14);
+        assert.equal(sut.abs(42.001), 42.001);
+    });
+});
```
lukehoban: Are these tests hooked up to a test harness?  Would be nice to have an `npm` script to run them - and to hook that into `make`.

sdk/nodejs/index.ts
jen20: I believe that is the behaviour that Terraform uses `filepath.basename` instead of `path.basename` in order to deal with Windows paths correctly. I added a test for this case to ensure it works with both types.

sdk/nodejs/index.ts
jen20: :+1:

sdk/nodejs/index.ts
jen20: Yes, not sure how that got past the compiler. Must be something named `list` in scope.

sdk/nodejs/tests/tests.ts
jen20: They are not currently, but will be prior to this PR being ready to merge.

sdk/nodejs/index.ts
jen20: :+1:

jen20:
@ellismg, as discussed, I've tagged you for review to ensure that the build system is correct. I've not _yet_ pulled in the package publishing script, but once the tests are passing in Travis CI, I will do so.

lukehoban:
This is quite stale now - so I'll close it out for now.  If/when we want to come back to it there's some great work in this branch.

Issue #183, opened 2018-06-19
 Python generated code should test against `six.text_types` for string type checks


swgillespie:
Generated code for Python tests against `basestring` for string type checks, which is not true for Python 3 strings.

swgillespie:
Fixed by https://github.com/pulumi/pulumi-terraform/pull/270.

Issue #184, opened 2018-06-19
Python generated code should use not use implicit relative imports anywhere


swgillespie:
See https://github.com/pulumi/pulumi/issues/1541. This is a Python 3 compat hazard.

yapith-suplari:
I got pass the relative import issues, then there's the basestring used everywhere issues. Then I wonder if someone is already tracking all the back-compat issues. Is there an umbrella issue that captures all kinds python 3 compat issues? I think it's useful to improve community visibility, so we can help track them and/or fix them.

swgillespie:
I've been tracking them here: https://github.com/pulumi/pulumi/issues/1535. I've also been using the `language/python3` tag to track individual issues.

swgillespie:
Fixed by https://github.com/pulumi/pulumi-terraform/pull/270.

Issue #168, opened 2018-06-07
Consider putting PEP-484 type hints on Python generated classes and functions 


swgillespie:
IDEs are pretty good at figuring out what types need to go where but they can be helped out significantly by the presence of PEP-484 type comments. We could significantly improve the Python IDE experience by doing this.

VS Code in particular is aware of these comments and will use them to provide improved code completion.

lukehoban:
@swgillespie We already implemented a fair bit of type hinting in the new Python3 pulumi SDK.  What is remaining at this stage?

swgillespie:
@lukehoban Python resources that tfgen emits look sort of like this:

```python
class SomeResource(pulumi.CustomResource):
    def __init__(self, name, props, opts=None):
        __props__ = {}
        __props__["foo"] = props["foo"]
        ...
        CustomResource.__init__(self, "foo:bar:baz", __props__, opts)
```

Ideally, we'd emit something like this:
```python
class SomeResource(pulumi.CustomResource):
    foo: Output[str] # <- note this
    """
    Docstring for Foo
    """

    def __init__(self, name, props, opts=None):
        __props__ = {}
        __props__["foo"] = props["foo"]
        ...
        CustomResource.__init__(self, "foo:bar:baz", __props__, opts)
```

The additional type hint at the root of the class greatly improves the IDE experience here. Also, as we venture into documentation territory for Python, we will need to place docstrings on the properties that we expose so that documentation generators can pick them up.

PR #270, opened 2018-10-31
Update codegen for Python 3

swgillespie:
Fixed a few Python 2/3 compataibility issues with tfgen generated code:
  1) Replace uses of basestring with str
  2) Replace all absolute imports with relative ones
  3) Remove all type checks for input properties
  4) Remove explicit setting of properties on self, now that the SDK
  does it

### DIFF
--- a/pkg/tfgen/generate_python.go
+++ b/pkg/tfgen/generate_python.go
@@ -230,7 +230,7 @@ func (g *pythonGenerator) emitIndex(mod *module, exports, submods []string) erro
 		}
 		w.Writefmtln("# Export this package's modules as members:")
 		for _, exp := range exports {
-			w.Writefmtln("from %s import *", exp)
+			w.Writefmtln("from .%s import *", exp)
 		}
 	}
 
@@ -432,7 +432,7 @@ func (g *pythonGenerator) emitResourceType(mod *module, res *resourceType) (stri
 	w.Writefmtln(`        """Create a %s resource with the given unique name, props, and options."""`, res.name)
 	w.Writefmtln("        if not __name__:")
 	w.Writefmtln("            raise TypeError('Missing resource name argument (for URN creation)')")
-	w.Writefmtln("        if not isinstance(__name__, basestring):")
+	w.Writefmtln("        if not isinstance(__name__, str):")
 	w.Writefmtln("            raise TypeError('Expected resource name to be a string')")
 	w.Writefmtln("        if __opts__ and not isinstance(__opts__, pulumi.ResourceOptions):")
 	w.Writefmtln("            raise TypeError('Expected resource options to be a ResourceOptions instance')")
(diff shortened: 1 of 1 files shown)


swgillespie:
Fixes https://github.com/pulumi/pulumi-terraform/issues/183, https://github.com/pulumi/pulumi-terraform/issues/184

swgillespie:
Note that this is a breaking change for Python and needs to be a part of a 0.17 release.

pgavlin:
> Note that this is a breaking change for Python and needs to be a part of a 0.17 release.

Can you elaborate on the breaks?

swgillespie:
> Can you elaborate on the breaks?

There are a few, but the upshot is that using a new package generated using this PR's tfgen will not be usable in tandem with downlevel Pulumi SDKs (<0.17). I'd express that dependency in `resources.go`, but we haven't produced a 0.17 SDK yet.

> Replace uses of basestring with str

This breaks anyone using Python 2. Since users could *only* use Python 2 until this morning, this breaks everybody.

> Replace all absolute imports with relative ones

This breaks anyone using Python 2 that don't have `from __future__ import absolute_imports`. Since the Python 2 SDK uses that, it should be fine.

> Remove explicit setting of properties on self, now that the SDK
does it

This breaks anyone not using the to-be-released 0.17 SDK (i.e. https://github.com/pulumi/pulumi/pull/2111).

So yeah. It's a hard break, but one we expected when transitioning to Python 3.

swgillespie:
ping - this look OK to merge?

pkg/tfgen/generate_python.go
```diff
@@ -446,30 +446,16 @@ func (g *pythonGenerator) emitResourceType(mod *module, res *resourceType) (stri
...
 		pname := pyName(prop.name)
-		ptype := pyType(prop)
 
 		// Fill in computed defaults for arguments.
 		if defaultValue := pyDefaultValue(prop); defaultValue != "" {
 			w.Writefmtln("        %s = %s", pname, defaultValue)
 		}
 
-		// Check that required arguments are present.  Also check that types are as expected.
+		// Check that required arguments are present.
```
joeduffy: Why did we remove the type assertions?

lukehoban:
BTW @swgillespie Do you have an example build of AWS with these changes?  Would be helpful to see the net impact of the changes in practice as well.

pkg/tfgen/generate_python.go
swgillespie: They aren't correct in the presence of `Future[T]` and `Output[T]` being used as inputs to resources, which we 1) now explicitly allow and 2) would fail all typechecks asserting that these types are exactly equal to a type T.

pkg/tfgen/generate_python.go
```diff
@@ -446,30 +446,16 @@ func (g *pythonGenerator) emitResourceType(mod *module, res *resourceType) (stri
...
-			w.Writefmt("        elif ")
-		} else {
-			w.Writefmt("        if %s and ", pname)
-		}
-		w.Writefmtln("not isinstance(%s, %s):", pname, ptype)
-		w.Writefmtln("            raise TypeError('Expected property %s to be a %s')", pname, ptype)
-
-		// Now perform the assignment, and follow it with a """ doc comment if there was one found.
-		w.Writefmtln("        __self__.%[1]s = %[1]s", pname)
-		if prop.doc != "" {
```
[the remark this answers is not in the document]
swgillespie: The correct place to put them now would be on PEP-484 type annotations on the class itself, which I hope to do in a follow-up PR for https://github.com/pulumi/pulumi-terraform/issues/168.

swgillespie:
Talked to @lukehoban offline and I think this is ready to merge if there aren't any further objections.

PR #275, opened 2018-11-06
Call Terraform Importer as part of Read operation

### LINKED ISSUE #266: Avoid needing to pass extra properties to `.get`
Today, `.get` takes an `id` and an optional bag of additional values.

When this is needed, it feels very awkward - it's both unclear why it is needed, and it is normally exactly duplicative of the `id`.

I believe that we might be able to get rid of the need for this entirely by invoking the `Importer` before the `Read`.  

For example, for `aws_lambda_function` - the `Importer` looks like:

```go
                Importer: &schema.ResourceImporter{
			State: func(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
				d.Set("function_name", d.Id())
				return []*schema.ResourceData{d}, nil
			},
		},
```

This is exactly the custom code we need to turn just an ID into a bag with enough properties to invoke `Read`.  I believe this is generally the contract for `Importer`.  As a result, the `tfbridge` code ideally invoke this prior to invoking the `Read` method.

jen20:
Terraform Importers exist to allow resources which cannot (or are not) read purely by ID to populate the necessary parts of configuration _from_ their ID. Previously, this functionality was unused in the Pulumi Terraform Bridge, leading to crashes (e.g. pulumi/pulumi-gcp#55) or inability to import without duplicating data points which form part of resource IDs (e.g. pulumi/pulumi-gcp#48).

This commit calls the importer as part of the read operation, if one is defined in the Terraform schema. It then adds the values set by the importer to the property bag passed to Read, iff they are unset already.

This fixes both classes of bugs - validation of IDs done in the importer now results in error messages being passed back in preference to panics, and the following program will now import a GCP resource:

```typescript
import * as gcp from "@pulumi/gcp";

const defaultPool = gcp.container.NodePool.get("default-pool",
     "us-central1-a/test-ci/default-pool");
// Note there is no redundant specification of data here.

export const nodePoolVersion = defaultPool.version;
```

Because an importer is capable of fanning out to multiple resource (for example, an AWS Security Group imports rules in addition to the group), it is necessary to match the returned data to the ID of the resource in order to determine a match.

Right now, we ignore nil states returned from the provider - while they are a Terraform provider bug if they occur, we have the ability to work around them by specifying properties directly, and it is not desirable to disable that ability at this stage by erroring.

There is some test coverage to show that this minimally works - it would be better to refactor some aspects of the provider to make it transport-agnostic to allow easier testing in future, as @joeduffy and I have discussed (and is tracked in #23).

I believe that plumbing this version through to the providers will:

- Fix #266 
- Fix pulumi/pulumi-gcp#55
- Fix pulumi/pulumi-gcp#48
- Lay groundwork for pulumi/pulumi#1635

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -59,6 +59,114 @@ type Resource struct {
 	TFName string           // the Terraform resource name.
 }
 
+// collectImporterAttributes runs the Terraform Importer defined on the Resource for the given
+// resource ID, and adds any non-ID attributes to the inputs map.
+func (res *Resource) collectImporterAttributes(resourceID resource.ID, inputs map[string]string) error {
+	// There is nothing to do here if the resource doesn't have an importer defined in the
+	// Terraform schema.
+	if res.TF.Importer == nil {
+		return nil
+	}
+
+	glog.V(9).Infof("%s has TF Importer", res.TFName)
+
+	id := resourceID.String()
+
+	// Prepare a Terraform ResourceData for the importer
+	data := res.TF.Data(nil)
+	data.SetId(id)
+	data.SetType(res.TFName)
+
+	// Run the importer defined in the Terraform resource schema
+	results, err := res.TF.Importer.State(data, nil)
+	if err != nil {
+		return errors.Wrapf(err, "importing %s", id)
+	}
+
+	// No resources were returned - error out
+	if len(results) < 1 {
+		return errors.Errorf("importer for %s returned no resources", id)
+	}
+
+	// Allow constructing an error in the case that we have a nil InstanceState returned from
+	// Terraform, which is always a programming error.
+	makeNilStateError := func(badResourceID string) error {
+		return errors.Errorf("importer for %s returned a empty resource state. This is always "+
+			"the result of a bug in the resource provider - please report this "+
+			"as a bug in the Pulumi provider repository.", badResourceID)
+	}
+
+	// A Terraform importer can return multiple ResourceData instances for different resources. For
+	// example, an AWS security group will also import the related security group rules as independent
+	// resources.
+	//
+	// Some Terraform importers _change_ the ID of the resource to allow for multiple formats to be
+	// specified by a user (for example, an AWS API Gateway Response). In the case that we only have
+	// a single ResourceData returned, we will use that ResourceData regardless of whether the ID
+	// matches, provided the resource Type does match.
+	//
+	// If we get multiple ResourceData back, we need to search the results for one which matches both
+	// the Type and ID of the resource we were trying to import (the "primary" InstanceState).
+	//
+	// The Type can be identified by looking at the ephemeral data attached to the InstanceState, since
+	// it is not stored in all cases - only for import.
+	var primaryInstanceState *terraform.InstanceState
+
+	if len(results) == 1 {
+		// Take the only result, assuming the Type matches
+		state := results[0].State()
+		if state == nil {
+			return makeNilStateError(id)
+		}
+		if state.Ephemeral.Type == res.TFName {
+			primaryInstanceState = state
+		}
+	} else {
+		// Search for a Type+ID match, and use the first (if any)
+		for _, result := range results {
+			if result.Id() != id {
+				continue
+			}
+
+			state := result.State()
+			if state == nil {
+				return makeNilStateError(id)
+			}
+
+			if state.Ephemeral.Type != res.TFName {
+				continue
+			}
+
+			primaryInstanceState = state
+			break
+		}
+	}
+
+	// No resources were returned - error out
+	if primaryInstanceState == nil {
+		return errors.Errorf("importer for %s returned no resources", id)
+	}
+
+	// Set any attributes the importer populated as values in the input map provided, in order
+	// that they are passed to the Read function. If any keys already exist, this is likely because
+	// they have been set in the `state` object passed to a `get` method, so we will honour those
+	// values by not overwriting existing keys.
+	for key, value := range primaryInstanceState.Attributes {
+		// ID is already set
+		if key == "id" {
+			continue
+		}
+
+		// Set the importer's value for they key if it is not already set
+		_, hasKey := inputs[key]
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/provider.go
```diff
@@ -59,6 +59,63 @@ type Resource struct {
...
+	data.SetId(id)
+	data.SetType(res.TFName)
+
+	results, err := res.TF.Importer.State(data, nil)
+	if err != nil {
+		return errors.Wrapf(err, "importing %s", id)
+	}
+
+	if len(results) < 1 {
+		return errors.Errorf("importer for %s returned no resources", id)
```
pgavlin: Any idea when this is possible with `err == nil`?

pkg/tfbridge/provider.go
```diff
@@ -59,6 +59,63 @@ type Resource struct {
...
+	}
+
+	// Find the resource with the matching ID, since an importer can return several (for example,
+	// a security group and associated rules), and set any non-id attributes as inputs to read.
+	for _, result := range results {
+		if result.Id() != id {
```
pgavlin: We'll want to check the type as well, right?

pkg/tfbridge/provider.go
```diff
@@ -59,6 +59,63 @@ type Resource struct {
...
+			return nil //TODO(jen20) Should we error here? Provider bug but will prevent workarounds
+		}
+
+		for key, value := range state.Attributes {
+			// ID is already set
+			if key == "id" {
+				continue
+			}
+
+			// Set the importer's value for they key if it is not already set
```
pgavlin: This is interesting. What are the tradeoffs here vs. just overwriting existing values? Are we leaving existing values to be consistent with user intent (which is good, I think)?

pkg/tfbridge/provider.go
jen20: Good catch, that needs adding too.

pkg/tfbridge/provider.go
jen20: Yes, the idea here was that you _can_ override things the importer might do in the `state` object to `get` calls if you want to.

pkg/tfbridge/provider.go
jen20: I don't _think_ it is possible, but I'm not sure relying on every importer behaving correctly here when the alternative is a panic is such a good idea?

pgavlin: >  the alternative is a panic is such a good idea

The alternative I was considering was just returning `nil` and letting `Refresh` do its thing. Not sure if that's better, though.

jen20: That could be a valid alternative - though if the resource is buggy enough that a defined importer doesn't return a state, I'm not sure what we can infer from that.

I'm happy to defer to your knowledge of the codebase here - which do you think is a better approach here given how we handle similar situations elsewhere?

pkg/tfbridge/provider.go
jen20: I've added a check for this after we verify that State is not nil, in b2b9be9.

jen20: It might be a good idea to restructure this code to separate out the search for the correct state from the action performed on it. I'll push another commit shortly.

pkg/tfbridge/provider.go
```diff
@@ -59,6 +59,70 @@ type Resource struct {
...
+
+	if len(results) < 1 {
+		return errors.Errorf("importer for %s returned no resources", id)
+	}
+
+	// Find the resource with the matching ID, since an importer can return several (for example,
+	// a security group and associated rules), and set any non-id attributes as inputs to read.
+	for _, result := range results {
+		if result.Id() != id {
+			continue
```
lukehoban: There are several Importers which change the ID.  This seems to be used as a way to allow multiple ID formats from users, while normalizing down to the one format understood by the provider.  This seems a useful thing to support from Pulumi in the same way (and users may expect it to work in the same way).

I think we should at least allow the case where only a single resource is returned to just assume that is the desired return value, and potentially also if there is a single resource of the desired type.

lukehoban: For example: https://github.com/terraform-providers/terraform-provider-aws/blob/c5f4c2c84362e81dcb9787aa1e2303891290c2cb/aws/resource_aws_api_gateway_gateway_response.go#L32.  But there are many others.

jen20: I think this is straightforward enough to support - I've taken a shot in b13b8a1.

jen20: For now I've erred on the side of caution and taken a single result of the correct type as likely being correct, but required a Type _and_ ID match if there are multiple resources returned. This is something I think we should revisit if it becomes an issue rather than relaxing immediately, though I'm open to alternative views!

pkg/tfbridge/provider.go
jen20: Offline discussion determined we should error here requesting a bug report in the same manner as Terraform does in the face of a nil state.

pkg/tfbridge/provider.go
```diff
@@ -59,6 +59,114 @@ type Resource struct {
...
+
+	id := resourceID.String()
+
+	// Prepare a Terraform ResourceData for the importer
+	data := res.TF.Data(nil)
+	data.SetId(id)
+	data.SetType(res.TFName)
+
+	// Run the importer defined in the Terraform resource schema
+	results, err := res.TF.Importer.State(data, nil)
```
lukehoban: Most of the other places we invoke resource providers we do so via the equivalent of `p.tf.ImportState`, which is effectively a wrapper over this.

I assume we're using the Importer directly here (and duplicating some of what `ImportState` does) because we want more control over some of the error cases?

jen20: Right - we could use `ImportState` but we'd have to check for non-concrete-type errors when there is no importer (which is ok with us). In addition, the error would direct people to report bugs in Terraform instead of Pulumi, which is probably not ideal since they are really our bugs to deal with.

PR #278, opened 2018-11-08
Remove vestigial generation of VERSION file

jen20:
This dates from before dependency information could easily be tracked either by a Go module file or the Dep lock, and is no longer needed.

Rather than update it to resolve version information from the Go module system and the local package cache, we remove the unnecessary file from generation.


joeduffy:
Good riddance! Thank you, LGTM 💯

Issue #276, opened 2018-11-08
Python: Properties come back from tfbridge in camel case


swgillespie:
Python would like all properties to be snake case by default. If Python code sends inputs to tfbridge in snake case, they will come back as camel case: https://github.com/pulumi/pulumi-terraform/blob/e75d240691570c101e6a87f8e974c81ba83492a1/pkg/tfbridge/names.go#L55-L95.

We need Python to observe the output properties with the same name that it saw them as inputs, so that it can resolve futures that it attached to the Resource object that was created.

PR #279, opened 2018-11-09
Emit snake case <-> camel case tables

swgillespie:
Newer versions of the Python SDK (https://github.com/pulumi/pulumi/pull/2187)
provide hooks for providers to translate to and from snake case and
camel case. This commit uses those hooks to query lookup tables at
runtime to convert between snake cased properties exposted to Python and
camel cased properties expected by tfbridge and providers in general.

Other half of https://github.com/pulumi/pulumi/pull/2187, fixes https://github.com/pulumi/pulumi-terraform/issues/276. See https://github.com/pulumi/pulumi-aws/compare/swgillespie/py3-tables for the diff of these changes applied to `pulumi-aws`.

Here are what the tables look like for `pulumi-aws`: https://github.com/pulumi/pulumi-aws/blob/37bc7483ee277728dda1e858481ae31ad6d975d3/sdk/python/pulumi_aws/tables.py.

### DIFF
--- a/pkg/tfgen/generate_python.go
+++ b/pkg/tfgen/generate_python.go
@@ -697,8 +697,8 @@ func (g *pythonGenerator) emitPropertyConversionTables(tableModule *module) erro
 	for key := range g.snakeCaseToCamelCase {
 		allKeys = append(allKeys, key)
 	}
-
 	sort.Strings(allKeys)
+
 	w.Writefmtln("_SNAKE_TO_CAMEL_CASE_TABLE = {")
 	for _, key := range allKeys {
 		value := g.snakeCaseToCamelCase[key]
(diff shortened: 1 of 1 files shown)


pkg/tfgen/generate_python.go
```diff
@@ -666,6 +684,96 @@ func (g *pythonGenerator) emitPackageMetadata(pack *pkg) error {
...
+		return err
+	}
+
+	defer contract.IgnoreClose(w)
+	var allKeys []string
+	for key := range g.snakeCaseToCamelCase {
+		allKeys = append(allKeys, key)
+	}
+
+	sort.Strings(allKeys)
```
pgavlin: Tiny nit: can you nudge this line up against the preceding loop? I very nearly missed it.

pkg/tfgen/generate_python.go
```diff
@@ -666,6 +684,96 @@ func (g *pythonGenerator) emitPackageMetadata(pack *pkg) error {
...
+	g.snakeCaseToCamelCase[snakeCaseName] = name
+	g.recordPropertyRec(sch, info)
+}
+
+// recordPropertyRec recurses through a property's schema and recursively records any properties contained within it.
+func (g *pythonGenerator) recordPropertyRec(sch *schema.Schema, info *tfbridge.SchemaInfo) {
+	switch sch.Type {
+	case schema.TypeList, schema.TypeSet:
+		// If this property that we are recursing on is a list or a set, we do not need to record any properties at this
+		// step but we do need to recurse into the element of the list or set.
```
pgavlin: nit: `s/element/element schema/` throughout

PR #280, opened 2018-11-12
Await pulumi.runtime.invoke when invoking

swgillespie:
`pulumi.runtime.invoke` is now asynchronous (https://github.com/pulumi/pulumi/pull/2175), so the wrappers that we generate around it must also be async and `await` the future returned from `pulumi.runtime.invoke`.


ellismg:
Going to merge this in and adopt it in the terraform providers.

Issue #281, opened 2018-11-12
Failure during `pulumi refresh` for `pulumi-random` when ingesting HEAD of pulumi-terraform


ellismg:
Trying to update the `pulumi-terraform` reference in `pulumi-random` and running into this problem during the refresh operation:

```
[ /pulumi-random/examples/simple ] Diagnostics:
[ /pulumi-random/examples/simple ]   random:index:RandomInteger: integer
[ /pulumi-random/examples/simple ]     error: Plan apply failed: importing 932: Invalid import usage: expecting {result},{min},{max} or {result},{min},{max},{seed}
```

https://travis-ci.com/pulumi/pulumi-random/builds/91115873#L1123 shows the failure.

lukehoban:
Added a longer note on https://github.com/pulumi/pulumi-terraform/pull/282#issuecomment-438322428, but I think this raises the issue that it's not necessarily safe to call `Importer` in `pulumi refresh` scenarios.

lukehoban:
We already fixed this - but by reverting #275.  This now tracks redoing #275 and avoiding this issue.

PR #282, opened 2018-11-13
Add ability to skip a Terraform importer

jen20:
During the implementation of #275, we made the assumption that it is always safe to run the Terraform importer specified for a resource as part of the read operation. During the rollout of this change to some ofthe more boutique providers (e.g. Random, and some resources in AWS), it was discovered by Matt that this does not hold, and that we introduce read errors because the format expected by the importer is not an ID or something which can be transformed into one.

This commit adds an option to the Pulumi schema for a resource to skip running the Terraform importer defined for that resource type, which should restore the old behaviour, while allowing the majority of resources for which running the importer is OK to retain their newer behaviour.

It may be the case in future that we want the ability to not just skip the Terraform importer but to define our own importer instead - if this is added, it should be considered that skipping the Terraform importer for resources with a Pulumi importer is desired, but this is a good first step.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -45,11 +45,23 @@ type ProviderInfo struct {
 // also give custom metadata for fields, using the SchemaInfo structure below.  Finally, a set of composite keys can be
 // given; this is used when Terraform needs more than just the ID to uniquely identify and query for a resource.
 type ResourceInfo struct {
-	Tok                 tokens.Type            // a type token to override the default; "" uses the default.
-	Fields              map[string]*SchemaInfo // a map of custom field names; if a type is missing, uses the default.
-	IDFields            []string               // an optional list of ID alias fields.
-	Docs                *DocInfo               // overrides for finding and mapping TF docs.
-	DeleteBeforeReplace bool                   // if true, Pulumi will delete before creating new replacement resources.
+	// a type token to override the default; "" uses the default.
+	Tok tokens.Type
+
+	// a map of custom field names; if a type is missing, uses the default.
+	Fields map[string]*SchemaInfo
+
+	// an optional list of ID alias fields.
+	IDFields []string
+
+	// Overrides for finding and mapping TF docs.
+	Docs *DocInfo
+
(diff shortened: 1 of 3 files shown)


jen20:
In conjunction with setting the new option in the Random provider, this should fix #281.

lukehoban:
This actually points out what is likely a larger problem with the Importer change, which I should have caught during review.

For a `.get` where we have just an `id` as input, calling `Importer` is absolutely what we want to do.  For a `pulumi refresh`, where we have what we believe is a fully-formed resources instance and just want to update it to the latest state, calling `Importer` is likely often going to be wrong.  It is certainly a use case that `Importers` have not been designed to handle, since in Terraform they are only ever called with an `id` as input.

I believe the problems with `Random` are because we are triggering calls to `Importer` during refresh.  I would not be surprised if there are many more similar issues.  This just isn't a scenario `Importer` is designed to be used in.

The problem is that the current Pulumi RPC model has only a single method here - `Read`, which gets used for both `.get()` and `pulumi refresh`.  That is sort of wrong.  `.get()` should be "Import then refresh" and `pulumi refresh` should be "just refresh".

I think there are a few options:
1. Try to manually annotate all the Importers that are unsafe to use during refresh - and ban them for both refresh and `.get`. (as in this PR)
2. Split the `Read` interface into separate `Import` and `Read` APIs and have `.get()` cause both to be invoked in order.
3. Change the current logic in `Read` to only invoke the `Importer` if `req.GetProperties()` is empty.  This should ensure that (a) `Import` gets called in any `.get()` scenarios (b) `Import` does not get called in cases where it will fail.

(3) may still not be quite enough though - cases where there are no input properties on a resource may still cause `refresh` to fail.

pgavlin:
After offline discussion with @lukehoban, option (3) seems a bit more appealing: it lets us solve this problem for TF-based providers specifically while we figure out if we do in fact need to adjust the model.

jen20:
Option 3 does seem like the best idea here - it was unclear how many importers are not suitable for use in the context of reading. I'll close this PR out and look at implementing the third option.

PR #284, opened 2018-11-15
Call Terraform Importer as part of get operations

jen20:
This is a reworking of commit b13b8a1b, which was merged into `pulumi-terraform` but subsequently reverted prior to release because of emergent bugs when refreshing some types of resources.

Terraform Importers exist to allow resources which cannot (or are not) read purely by ID to populate the necessary parts of configuration _from_ their ID. Previously, this functionality was unused in the Pulumi Terraform Bridge, leading to crashes (e.g. pulumi/pulumi-gcp#55) or inability to import without duplicating data points which form part of resource IDs (e.g. pulumi/pulumi-gcp#48).

This commit calls the importer as part of the read operation, provided both of the following conditions are true:

- an importer is defined in the Terraform schema
- the properties map on the Read request is empty, indicating a get operation instead of a refresh operation

After running the importer, we add the values set by the importer to the property bag passed to Read.

This fixes both classes of bugs - validation of IDs done in the importer now results in error messages being passed back in preference to panics, and the following program will now import a GCP resource:

```typescript
import * as gcp from "@pulumi/gcp";

const defaultPool = gcp.container.NodePool.get("default-pool",
     "us-central1-a/test-ci/default-pool");
// Note there is no redundant specification of data here.

export const nodePoolVersion = defaultPool.version;
```

Because an importer is capable of fanning out to multiple resource (for example, an AWS Security Group imports rules in addition to the group), it is necessary to match the returned data to the ID of the resource in order to determine a match.

Right now, we ignore nil states returned from the provider - while they are a Terraform provider bug if they occur, we have the ability to work around them by specifying properties directly, and it is not desirable to disable that ability at this stage by erroring.

There is some test coverage to show that this minimally works - it would be better to refactor some aspects of the provider to make it transport-agnostic to allow easier testing in future.

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -59,6 +59,108 @@ type Resource struct {
 	TFName string           // the Terraform resource name.
 }
 
+// runTerraformImporter runs the Terraform Importer defined on the Resource for the given
+// resource ID, and returns a replacement input map if any resources are matched. A nil map
+// with no error should be interpreted by the caller as meaning the resource does not exist,
+// but there were no errors in determining this.
+func (res *Resource) runTerraformImporter(resourceID resource.ID) (map[string]string, error) {
+	// There is nothing to do here if the resource doesn't have an importer defined in the
+	// Terraform schema.
+	if res.TF.Importer == nil {
+		return nil, nil
+	}
+
+	glog.V(9).Infof("%s has TF Importer", res.TFName)
+
+	id := resourceID.String()
+
+	// Prepare a Terraform ResourceData for the importer
+	data := res.TF.Data(nil)
+	data.SetId(id)
+	data.SetType(res.TFName)
+
+	// Run the importer defined in the Terraform resource schema
+	results, err := res.TF.Importer.State(data, nil)
+	if err != nil {
+		return nil, errors.Wrapf(err, "importing %s", id)
+	}
+
+	// No resources were returned. There are a few different ways this can happen - principally
+	//  - The resource never existed
+	//  - The resource did exist but was deleted
+	//
+	// The engine is capable of converting an empty response into an appropriate error for the
+	// user, so we don't want to disable that behaviour by returning our own (likely different)
+	// error up the chain. Instead, we return a nil map _and_ a nil error, and it is the
+	// responsibility of the caller to convert this into an appropriate error message.
+	//
+	// We consider the case in which multiple results are returned from the importer, but none
+	// match the ID expected to be an error, and this is handled later in this function.
+	if len(results) < 1 {
+		return nil, nil
+	}
+
+	// Allow constructing an error in the case that we have a nil InstanceState returned from
+	// Terraform, which is always a programming error.
+	makeNilStateError := func(badResourceID string) error {
+		return errors.Errorf("importer for %s returned a empty resource state. This is always "+
+			"the result of a bug in the resource provider - please report this "+
+			"as a bug in the Pulumi provider repository.", badResourceID)
+	}
+
+	// A Terraform importer can return multiple ResourceData instances for different resources. For
+	// example, an AWS security group will also import the related security group rules as independent
+	// resources.
+	//
+	// Some Terraform importers _change_ the ID of the resource to allow for multiple formats to be
+	// specified by a user (for example, an AWS API Gateway Response). In the case that we only have
+	// a single ResourceData returned, we will use that ResourceData regardless of whether the ID
+	// matches, provided the resource Type does match.
+	//
+	// If we get multiple ResourceData back, we need to search the results for one which matches both
+	// the Type and ID of the resource we were trying to import (the "primary" InstanceState).
+	//
+	// The Type can be identified by looking at the ephemeral data attached to the InstanceState, since
+	// it is not stored in all cases - only for import.
+	var primaryInstanceState *terraform.InstanceState
+
+	if len(results) == 1 {
+		// Take the only result, assuming the Type matches
+		state := results[0].State()
+		if state == nil {
+			return nil, makeNilStateError(id)
+		}
+		if state.Ephemeral.Type == res.TFName {
+			primaryInstanceState = state
+		}
+	} else {
+		// Search for a Type+ID match, and use the first (if any)
+		for _, result := range results {
+			if result.Id() != id {
+				continue
+			}
+
+			state := result.State()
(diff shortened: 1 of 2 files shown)


joeduffy:
Sorry for the delay looking at this. @pgavlin, will you have a chance to look soon? I'll give it a look but I'm far less qualified.

pkg/tfbridge/provider.go
```diff
@@ -59,6 +59,106 @@ type Resource struct {
...
+	}
+
+	// No resources were returned - error out
+	if primaryInstanceState == nil {
+		return errors.Errorf("importer for %s returned no resources", id)
+	}
+
+	// Set any attributes the importer populated as values in the input map provided, with the exception of ID
+	// which should already be set.
+	for key, value := range primaryInstanceState.Attributes {
```
lukehoban: I think in the modified approach here, there will never be any `inputs` to this function, and we should not have logic for merging the import with the `inputs`.  I assume we can remove this code and just return the imported values directly?

pkg/tfbridge/provider.go
```diff
@@ -59,6 +59,106 @@ type Resource struct {
...
+	// Run the importer defined in the Terraform resource schema
+	results, err := res.TF.Importer.State(data, nil)
+	if err != nil {
+		return errors.Wrapf(err, "importing %s", id)
+	}
+
+	// No resources were returned - error out
+	if len(results) < 1 {
+		return errors.Errorf("importer for %s returned no resources", id)
+	}
```
lukehoban: Nit - and it may be cleaner the way you have it - but I believe this condition is redundant with the later check that returns the same error.

jen20: It is redundant (and vestigial of the previous pass at this pull request). I’m not sure it’s necessarily worse, though these are two different error conditions:

- Here, the Terraform importer returned no results at all
- In the repeated check later, there may have been several results from the Terraform importer but none satisfied our criteria for matching the imported resource.

Depending on whether we want to watch out explicitly for that condition, it might be better to make the error message reflect the different nature of the failure to make identifying any issues which come up in the wild a bit easier? If we don’t care about distinguishing between these conditions, we can remove this check as you suggest.

pkg/tfbridge/provider.go
jen20: Yes, you’re right. I’ll make this change.

pkg/tfbridge/provider.go
jen20: In the latest version of this I’ve opted to keep the additional check, but change the wording of the later path to indicate that there were no matching results rather than no results returned.

jen20:
@lukehoban Thanks for the review, I’ve pushed a new version of this which addresses the feedback. Since it makes a couple of changes to the structure I’ll wait for you to take another look before merging.

pkg/tfbridge/provider.go
```diff
@@ -59,6 +59,96 @@ type Resource struct {
...
+	data.SetType(res.TFName)
+
+	// Run the importer defined in the Terraform resource schema
+	results, err := res.TF.Importer.State(data, nil)
+	if err != nil {
+		return nil, errors.Wrapf(err, "importing %s", id)
+	}
+
+	// No resources were returned - error out
+	if len(results) < 1 {
```
pgavlin: Just out of curiosity, what happens if the importer is called for a resource that has been deleted out-of-band? Could it legitimately return 0 resources?

pkg/tfbridge/provider.go
```diff
@@ -59,6 +59,96 @@ type Resource struct {
...
+		if state == nil {
+			return nil, makeNilStateError(id)
+		}
+		if state.Ephemeral.Type == res.TFName {
+			primaryInstanceState = state
+		}
+	} else {
+		// Search for a Type+ID match, and use the first (if any)
+		for _, result := range results {
+			if result.Id() != id {
```
pgavlin: I thought I remembered cases where we've seen ID mismatches between the input ID and the imported ID. Should we call out that possibility here?

jen20: We decided the last time around that if this was the case and only one resource was returned, we would use that resource, but otherwise we considered this irreconcilable.

pkg/tfbridge/provider.go
jen20: Would get not fail in this case anyway during refresh?

pgavlin: No: it would return a `nil` state, which the engine recognizes as “this resource was deleted.”

pkg/tfbridge/provider.go
pgavlin: Can we leave a comment to that effect?

pkg/tfbridge/provider.go
jen20: Ah, then it is a possibility. What would we expect to happen if a resource we tried to get were deleted or never existed? Presumably continuing blind would just result in further errors?

pgavlin: IIRC the engine or language provider will turn a missing resource during a `get` into an error.

jen20: I've modified the importer code to respect this behaviour and use the engine behaviour for dealing with an empty state from import.

pkg/tfbridge/provider.go
jen20: I've added a much longer form comment around the behaviour and rationale above this block.

jen20:
@lukehoban I think this can be merged following the last round of changes, pending CI. The downstream effect seems most noticeable in the AWS provider which has some kind of build failure, though it does not look related to this.

jen20:
The AWS downstream is broken but not in a way which seems relevant to this. Now all the feedback has been addressed I’ll go ahead and merge, and open PRs downstream to pull in this version and update providers.

jen20:
The thorn in the side of that plan is that I can't merge this (via GitHub at least) without the downstream being fixed already since the AWS build is failing for what looks to me like unrelated reasons. I'll try pulling this down locally to see what the issue is.

jen20:
The failure _is_ related to this PR, but not to the code contained in it - our method of testing downstream impact isn't valid for AWS anymore. This is now tracked in #286.

lukehoban:
The failing test is blocked on https://github.com/pulumi/pulumi-terraform/issues/286.  @jen20 verified manually that the AWS package does build correctly with these changes, so overriding GitHub checks and merging this.

PR #290, opened 2018-11-30
Accept CustomResourceOptions in `get` methods

### LINKED ISSUE #289: Accept `CustomResourceOptions` in generated `get` methods
e.g. here is the `get` method for `aws.s3.Bucket`:

```
    public static get(name: string, id: pulumi.Input<pulumi.ID>, state?: BucketState): Bucket {
        return new Bucket(name, <any>state, { id });
    }
```

Note that it does not accept `CustomResourceOptions`, and thus cannot be used with a first-class provider. This should be a simple fix.

pgavlin:
This is necessary in order to use first-class providers with these
methods.

Fixes #289.

### DIFF
--- a/pkg/tfgen/generate_nodejs.go
+++ b/pkg/tfgen/generate_nodejs.go
@@ -439,6 +439,7 @@ func (g *nodeJSGenerator) emitPlainOldType(w *tools.GenWriter, pot *plainOldType
 	w.Writefmtln("}")
 }
 
+//nolint:lll
 func (g *nodeJSGenerator) emitResourceType(mod *module, res *resourceType) (string, error) {
 	// Create a resource module file into which all of this resource's types will go.
 	name := res.name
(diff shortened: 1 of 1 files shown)


joeduffy:
LGTM.

Only question is whether we should generate an overload

```
public static get(
    name: string, id: pulumi.Input<pulumi.ID>, opts?: pulumi.CustomResourceOptions)
```

since it seems likely to me that `opts` will be a far common parameter to pass than `state`.

pgavlin:
> Only question is whether we should generate an overload

We can generate that overload, but I'm not sure about the usability. The only way we can differentiate the `state` bag from the `options` bag is by the properties each contain, and there's no guarantee that there won't be overlap. If there is overlap, we may put users in a confusing situation in which they seem to be passing a `state` bag but it is being silently interpreted as an `options` bag. They could resolve this by using the 4-argument overload.

Thoughts?

lukehoban:
> We can generate that overload, but I'm not sure about the usability. The only way we can differentiate the state bag from the options bag is by the properties each contain, and there's no guarantee that there won't be overlap. If there is overlap, we may put users in a confusing situation in which they seem to be passing a state bag but it is being silently interpreted as an options bag. They could resolve this by using the 4-argument overload.

Yeah - I was thinking about that as well - and I can't think of a solution that feels "good".  And just allowing these to overlap feels like it will be troublesome for adding fields to the `options` bag in the future.

Although I agree with @joeduffy's concern about the common case looking less nice here, I think it's not bad enough to motivate doing something overly awkward and compatibility-fragile to fix, at least for now.

I'd suggest we take this PR as is.

PR #294, opened 2018-12-10
Pass TF metadata to resource importers

pgavlin:
This is required for certain importers to operate (e.g. the importer for
an S3 bucket).

Fixes https://github.com/pulumi/pulumi-aws/issues/394.


chrsmith:
The code seems reasonable. Is it possible to add a regression test for this? Or perhaps if not in this repo, in `pulumi-aws`?

jen20:
I didn’t realise that importers were using metadata in the way - this looks like a good fix.

jen20:
@chrsmith I’ll add a test to the AWS provider based on your failing example.

PR #295, opened 2018-12-11
Converge lock file with dep v0.5.0

jen20:
The Gopkg.lock file was out of sync with the inputs for Gopkg.toml for dep v0.5.0. This commit adds the converged lock file.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,35 +2,46 @@
 
 
 [[projects]]
+  digest = "1:61c24bdb660a326b420d5422c4fd124041b4aac88e212c6ac75f3a57b0b9cada"
   name = "github.com/agext/levenshtein"
   packages = ["."]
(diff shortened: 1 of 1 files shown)


pgavlin:
Do we need this? We're still using dep 0.4.x in CI, etc.

jen20:
Actually no - for some reason I was under the impression that @ellismg's recent work had updated dep to 0.5.0, but apparently not.

ellismg:
> Actually no - for some reason I was under the impression that @ellismg's recent work had updated dep to 0.5.0, but apparently not.

At the time, it looked like it was likely we were going to land on modules, so I didn't want to churn all of our Gopkg.lock files so we could update to 0.5.0 just to delete it in a week or so.

I'm happy to move us forward (across all repositories) if we think we are going to be on dep for a while.

PR #299, opened 2018-12-29
Convert HCL examples to Pulumi during code-gen

joeduffy:


Gopkg.template.toml
```diff
@@ -0,0 +1,28 @@
+[[constraint]]
```
lukehoban: Are these Gopkg changes needed as part of this change?

joeduffy: No, I plan to remove them before marging, although I definitely recommend trying to eliminate the use of our private Terraform fork as a separate change.

PR #303, opened 2019-01-11
Stop cleaning Terraform schema in tfbridge

jen20:
Cleaning the Terraform schema using CleanTerraformSchema results in properties which are marked Removed by the schema definition being actually removed. Such properties can be set by CRUD methods for a Terraform resource, however, so we can't just remove them.

By not cleaning in tfbridge but still retaining the calls to Clean in tfgen, we should remove the class of errors seen in pulumi/pulumi-gcp#79. This approach does mean that items which are Removed but still set are saved in state, but that is not likely to present an issue.

An alternative is to filter the removed fields _after_ the call to the Terraform provider but before returning the new state to the engine - though this could present issues if a removed field is used as an input. We aren't aware of cases where this happens, but cannot necessarily rule it out.

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -177,7 +177,7 @@ func NewProvider(ctx context.Context, host *provider.HostClient, module string, 
 		version: version,
 		tf:      tf,
 		info:    info,
-		config:  CleanTerraformSchema(tf.Schema),
+		config:  tf.Schema,
 	}
 	p.setLoggingContext(ctx)
 	p.initResourceMaps()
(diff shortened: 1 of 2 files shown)


lukehoban:
> This approach does mean that items which are Removed but still set are saved in state, but that is not likely to present an issue.

I'm trying to understand what would happen here - and what kind of problem it could theoretically present.  I suppose this means that the JS caller of `registerResource` may get back some of these properties, even though they don't have actual properties on the JS object for them.  I imagine we already handle that correctly as a result of having to support imprecise plugin versions?  Any other scenarios to be potentially concerned about?

jen20:
@lukehoban that was the one that @pgavlin and I could think of when we discussed this earlier.

lukehoban:
We should try pushing this out to at least one of the providers and get some validation on it in provider repos (plus an examples nightly?) so that we can get out a fix soon for https://github.com/pulumi/pulumi-gcp/issues/79 which has hit several users of the latest GCP provider.

PR #304, opened 2019-01-15
Improve docstring generation for Python

swgillespie:
Improve the docstrings that we generate for Python classes so that
documentation generators like Sphinx can make sense of them. This commit
places docstrings on all output properties and a docstring on `__init__`
outlining the inputs of the resource.

### DIFF
--- a/pkg/tfgen/generate_python.go
+++ b/pkg/tfgen/generate_python.go
@@ -15,6 +15,7 @@
 package tfgen
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -431,9 +432,7 @@ func (g *pythonGenerator) emitResourceType(mod *module, res *resourceType) (stri
 	}
 	// Produce a class definition with optional """ comment.
 	w.Writefmtln("class %s(%s):", pyClassName(res.name), baseType)
-	if res.doc != "" {
-		g.emitDocComment(w, res.doc, "    ")
-	}
+	g.emitMembers(w, mod, res)
 
 	// Now generate an initializer with arguments for all input properties.
 	w.Writefmt("    def __init__(__self__, __name__, __opts__=None")
@@ -444,7 +443,7 @@ func (g *pythonGenerator) emitResourceType(mod *module, res *resourceType) (stri
 	}
 	w.Writefmtln("):")
 
-	w.Writefmtln(`        """Create a %s resource with the given unique name, props, and options."""`, res.name)
+	g.emitInitDocstring(w, mod, res)
 	w.Writefmtln("        if not __name__:")
 	w.Writefmtln("            raise TypeError('Missing resource name argument (for URN creation)')")
 	w.Writefmtln("        if not isinstance(__name__, str):")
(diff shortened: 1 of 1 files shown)


swgillespie:
Fixes https://github.com/pulumi/pulumi-terraform/issues/168 but more concretely makes it much easier to generate documentation from tfgen-generated code.

pkg/tfgen/generate_python.go
```diff
@@ -794,6 +793,82 @@ func (g *pythonGenerator) recordPropertyRec(sch *schema.Schema, info *tfbridge.S
...
 
+// emitMembers emits property declarations and docstrings for all output properties of the given resource.
+func (g *pythonGenerator) emitMembers(w *tools.GenWriter, mod *module, res *resourceType) {
+	for _, prop := range res.outprops {
+		name := pyName(prop.name)
+		ty := pyType(prop)
+		w.Writefmtln("    %s: pulumi.Output[%s]", name, ty)
+		if prop.doc != "" {
+			g.emitDocComment(w, prop.doc, "    ")
+		}
```
lukehoban: I had no idea this was a standard format for property documentation - and couldn't find any examples of this in the wild (and VS Code doesn't seem to pick up the doc strings?).  But I assume this works for Sphinx docgen?

pkg/tfgen/generate_python.go
```diff
@@ -794,6 +793,82 @@ func (g *pythonGenerator) recordPropertyRec(sch *schema.Schema, info *tfbridge.S
...
+	}
+}
+
+// emitInitDocstring emits the docstring for the __init__ method of the given resource type.
+//
+// Sphinx (the documentation generator that we use to generate Python docs) does not draw a distinction between
+// documentation comments on the class itself and documentation comments on the __init__ method of a class. The docs
+// repo instructs Sphinx to concatenate the two together, which means that we don't need to emit docstrings on the class
+// at all as long as the __init__ docstring is good enough.
```
lukehoban: One note - with this new approach VS Code no longer renders the help text for the class when hovering over the class name.

It does render the concatenation of class and constructor help text as parameter help on the constructor call.

Should we split these up and just let Sphinx re-concatenate them?  Seems that would result in good experience for both, and a generally more readable documentation in the code itself?

swgillespie: We *could* actually put the docstring in both locations and then tell sphinx to ignore the class docstring - I think that would work.

lukehoban: I actually sort of like keeping the class doc comment and the constructor doc comment separate.  Feels like there's a pretty natural way to do that (and it's the same we do for Node.js).  And it seems both Sphinx and VSCode can handle this well.  Any reason not to do that?

pkg/tfgen/generate_python.go
swgillespie: It does. This is the standard Sphinx autodoc form. There are a few other forms out there (NumPy and Google both have their own forms) but as far as I can tell this is "the standard".

The fact that VSCode's Python extension doesn't handle this well seems to be well-documented: https://github.com/Microsoft/vscode-python/issues/38. It's pretty unfortunate, if this issue is still relevant.

swgillespie: Based on the preponderance of tensorflow/scipy, I do wonder if numpy/google's style is "winning", but at this point all we really need to do is generate something that Sphinx can understand, and Sphinx can understand any of those with the right config.

pkg/tfgen/generate_python.go
swgillespie: I'll have to see what Sphinx does with `__init__` parameters - I'll try it today.

pkg/tfgen/generate_python.go
lukehoban: Yep - makes sense - totally fine with this as-is.

pkg/tfgen/generate_python.go
swgillespie: @lukehoban I tried this today and Sphinx did not handle this very well - it re-rendered the class docstring as expected but it mangled the `__init__` docstring pretty badly.

I think I'd rather merge what we've got now and iterate on the dev experience as we go, is that OK?

PR #309, opened 2019-01-18
Update pulumi/pulumi dependency

ellismg:
Pick up the fix for https://github.com/pulumi/pulumi/pull/2368. At a
minimum, this version of pulumi/pulumi needs to flow to each of our
tf-bridged providers, and it seems like because of that we should also
update our reference here.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -412,7 +412,7 @@
     "pkg/workspace",
     "sdk/proto/go"
   ]
-  revision = "0e868f15fc3aad99ac77eca29f8c0600a27b5ae3"
+  revision = "a02bfb64696d90bc8a5491b718f1f6853f041e24"
 
 [[projects]]
   branch = "master"


ellismg:
@jen20 I need to go ahead and ensure that at least this version of pulumi/pulumi is used in all our resource providers.  What's the best way for me to do that?  I know you are working on pushing terraform updates to other repositories.  Should I land this and then update aws and let you handle the rest as you look at the examples generation issues?

jen20:
@ellismg I think once #310 is merged, we should be able to update all of the tfbridge repos without issue.

ellismg:
> @ellismg I think once #310 is merged, we should be able to update all of the tfbridge repos without issue.

Are you OK with me merging this now so it can go along for the ride with #310?

jen20:
Yup, I can roll it out to the other repos.

Issue #306, opened 2019-01-16
Push out examples code generation to all repos


lukehoban:
We added the ability to generate examples as part of doc generation for tfgen.  We should push this out to all the tfgen-based repos.

jen20:
As part of this, we want to hard fail the build if `tf2pulumi` is not available, in order to prevent #308.

jen20:
This is now complete.

PR #310, opened 2019-01-18
Add -allow-missing-plugins to tf2pulumi calls

jen20:
This contributes to fixing #306.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -395,7 +395,7 @@ func convertHCL(hcl string) (string, string, error) {
 	// tf2pulumi in library form because it greatly complicates our modules/vendoring story.
 	stdout := &bytes.Buffer{}
 	stderr := &bytes.Buffer{}
-	tf2pulumi := exec.Command(path)
+	tf2pulumi := exec.Command(path, "-allow-missing-plugins")
 	tf2pulumi.Dir = dir
 	tf2pulumi.Stdout = stdout
 	tf2pulumi.Stderr = stderr


joeduffy:
I am not an expert on this, so will want @pat to weigh in.

My only worry here is that this might lead to different code-gen locally versus on Travis, still, because when the plugin is missing we won't use our name-mappings from the plugin. I think this might lead to names like `some_random_property`, rather than `someRandomProperty`.

Unfortunately, I don't have a great "alternative" to suggest. One idea is to pass an explicit set of providers at the command line to use, and tell `tf2pulumi` to skip anything outside of that set. Then for, say, Azure, we'd pass `--plugins=/path/to/pulumi-resource-azure`, and that's all that it would use.

pgavlin:
> Unfortunately, I don't have a great "alternative" to suggest. One idea is to pass an explicit set of providers at the command line to use, and tell tf2pulumi to skip anything outside of that set.

I think that we should do this.

We could also have Travis install a select set of basic plugins (`random` and `template` come to mind) if the examples that fail generation due to those plugins are high-value.

lukehoban:
>  degrades the generated code

Does it really degrade the generated code in these cases, or just allow examples which are "incomplete" to still codegen?

joeduffy:
Won't it still be different on dev versus Travis machines? Uses the plugin on one, degraded code on the other? We may get lucky for the current example set but it would seem this is a lurking issue...

pgavlin:
> Does it really degrade the generated code in these cases, or just allow examples which are "incomplete" to still codegen?

It does degrade the code: any uses of resources managed by a missing provider will look a bit weird due to missing type information. I don't have an example on hand, but names especially can be strange.

> Won't it still be different on dev versus Travis machines? Uses the plugin on one, degraded code on the other?

Yes, it will. I do think that we should do the thing you suggested earlier and add a command-line flag to limit the plugins available to `tf2pulumi`.

joeduffy:
LGTM. Now that we have a workaround for the plugin references, are we good to merge this?

Once we do, I'll re-gen the docs website.

jen20:
We came up with an alternative to merging that, which is to install the necessary plugins in the Makefile of each particular provider. I think it's better to fail the build as new providers are required and for us to update the Makefile in response vs allowing worse docs to slide through.

PR #313, opened 2019-01-22
Update Gopkg.lock for dep 0.5

jen20:


jen20:
Restarting the build here because the downstream impact tests didn't take into account the latest code.

joeduffy:
I did this in my recent commit, so closing out.

PR #314, opened 2019-02-01
Parse more examples

joeduffy:
We previously only tried to parse examples that were obviously inside
of "examples" fragments in the source Markdown. Unfortunately, this
meant we skipped a whole lot of HCL, since it shows up in other places
like front matter sections, inconsistently labeled H2s and H3s, etc.

This change just attempts to parse it all. If we encounter a failure,
we still keep parsing away, and it's up to the caller of parseExamples
to figure out how to proceed. The heuristic I landed on after much
experimentation is to elide entire example sections, but to just keep
the code sample out of the resulting docs for essential sections; this
ensures that we don't skip entire front matter, argument, or attribute
sections, while not ended up with blank example sections that look
funny. This heuristic isn't perfect, however it seems to fare well in
practice across the AWS and Azure providers, at least.

The result is looking pretty great!

This fixes pulumi/docs#784.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,35 +2,46 @@
 
 
 [[projects]]
+  digest = "1:61c24bdb660a326b420d5422c4fd124041b4aac88e212c6ac75f3a57b0b9cada"
   name = "github.com/agext/levenshtein"
   packages = ["."]
+  pruneopts = ""
   revision = "5f10fee965225ac1eecdc234c09daf5cd9e7f7b6"
   version = "v1.2.1"
 
 [[projects]]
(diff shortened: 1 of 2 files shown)


pkg/tfgen/generate.go
```diff
@@ -370,7 +370,7 @@ func (g *generator) Generate() error {
 	}
 
 	// Print out some documentation stats as a summary afterwards.
-	printDocStats(false, false)
+	printDocStats(true, true)
```
pgavlin: Do we want this debugging info by default now?

joeduffy: Oops, unintentional! Will fix

PR #320, opened 2019-02-04
Change name of Python name and opts arguments

swgillespie:
The given names of `__name__` and `__opts__` are valid Markdown, which
confuses our documentation generator and causes us to generate extremely
poor Markdown-flavored documentation. This PR changes the default name
to "resource_name" and the option bag to "opts".

This is a breaking change, so for backwards compatability this change
appends `__name__` and `__opts__` arguments to the end of every resource's
constructor that aliases resource_name and opts. If a user uses those
parameters, their code works the same as before, except that they will
receive a deprecation warning if they have enabled warnings. `__name__`
and `__opts__` are not documented (so they don't break the docs generator)
but they will continue to work until they are removed in a future
release.

### DIFF
--- a/pkg/tfgen/generate_python.go
+++ b/pkg/tfgen/generate_python.go
@@ -104,6 +104,7 @@ func (g *pythonGenerator) openWriter(mod *module, name string, needsSDK bool) (*
 
 func (g *pythonGenerator) emitSDKImport(mod *module, w *tools.GenWriter) {
 	w.Writefmtln("import json")
+	w.Writefmtln("import warnings")
 	w.Writefmtln("import pulumi")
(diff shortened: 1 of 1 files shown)


swgillespie:
View of the deprecation warning:

screen shot 2019-02-04 at 1 08 23 pm

Users must opt-in to warnings, so they will not see this by default unless they do this:

```python
import warnings
warnings.simplefilter('always')
```

lukehoban:
Curious - what was the original motivation for these `__` names?

swgillespie:
My best guess is to avoid unintentional collision with property names.

swgillespie:
ping - I'm going to merge this today if there aren't any objections.

Issue #319, opened 2019-02-04
Python presence checks for required fields are incorrect


swgillespie:
Python codegen uses the truthiness of a particular value to test whether or not it was provided:

https://github.com/pulumi/pulumi-random/blob/2e26200bd1db7828bfcc28c9531dc868de20013f/sdk/python/pulumi_random/random_integer.py#L63-L65

This isn't correct in general, since there are many provided values of a property that are not true but are present (for example, the number `0`).

We should instead be testing with `if prop is not None`.

swgillespie:
This has been fixed.

PR #321, opened 2019-02-04
Emit correct presence checks for required props

swgillespie:
Coercing a value to a boolean is not enough to determine whether a
property was provided in Python - it may have been provided, but it may
be a value that Python does not consider to be truthy, such as the
number zero.

This commit tests properties explicitly against None (their Python
default value) to check and see if they have been provided.


swgillespie:
Fixes https://github.com/pulumi/pulumi-terraform/issues/319.

swgillespie:
ping - I'm going to merge this today if there aren't any objections.

PR #322, opened 2019-02-08
Optionally elide name properties from examples

pgavlin:
For most TF-based providers, we have a fairly regular pattern of
auto-generating values for resource names if they are not specified.
This helps avoid collisions between resources, and increases the chance
that copy-pasted code will deploy successfully. These changes ask
`tf2pulumi` to strip any values for autoname properties from its output
when generating examples.


pgavlin:
Note that this depends on https://github.com/pulumi/tf2pulumi/pull/66 being merged, released, and rolled out via https://github.com/pulumi/scripts/blob/master/ci/install-common-toolchain.sh.

PR #330, opened 2019-02-19
Fall back to the schema type for strings

swgillespie:
Terraform occasionally sends back strings for fields whose schema type
indicates that they are not strings. If this is the case, convert the
string to the requested primitive instead of returning the string
verbatim.

This fixes a variety of type errors that can occur when using resource
outputs in strongly typed languages.

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -477,43 +477,7 @@ func MakeTerraformOutput(v interface{},
 	case reflect.Float64:
 		return resource.NewNumberProperty(val.Float())
 	case reflect.String:
-		// If the string is the special unknown property sentinel, reflect back an unknown computed property.  Note that
-		// Terraform doesn't carry the types along with it, so the best we can do is give back a computed string.
-		t := val.String()
-		if t == config.UnknownVariableValue {
-			return resource.NewComputedProperty(
-				resource.Computed{Element: resource.NewStringProperty("")})
-		}
-
-		// It's just a string. It might be another value, depending on the schema. Do we have a schema available?
-		if tfs == nil {
-			// If not, it's probably an ID or some other Pulumi-specific output. Leave it alone.
-			return resource.NewStringProperty(t)
-		}
-
-		switch tfs.Type {
-		case schema.TypeInt:
-			intVal, err := strconv.ParseInt(t, 0, 0)
-			if err != nil {
-				return resource.NewStringProperty(t)
-			}
-			return resource.NewNumberProperty(float64(intVal))
-		case schema.TypeBool:
-			boolVal, err := strconv.ParseBool(t)
-			if err != nil {
-				return resource.NewStringProperty(t)
-			}
-			return resource.NewBoolProperty(boolVal)
-		case schema.TypeFloat:
-			floatVal, err := strconv.ParseFloat(t, 64)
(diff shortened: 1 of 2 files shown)


swgillespie:
Fixes https://github.com/pulumi/pulumi/issues/2459.

I think that this has been OK up until this point because JavaScript plays very fast and loose with types and freely coerces them to be whatever they need to be, and also because Pulumi programs in general don't often manipulate integers.

I ran into this when working with Python because Python is much less willing to coerce between primitive types. This PR applies a pretty simple transformation to output properties: If a property's reflected type is `string`, and there is a schema present that indicates that the property is actually some other type (bool, float, or int), parse the string and return the requested primitive type. If any parse fails, or if there is no schema present, the string field is left as-is. I'm tagging the experts for review since I am not 100% sure of the full implications of this.

swgillespie:
> To that end, would you mind pulling the coercion logic out into an exported function that `tf2pulumi` can consume?

Will do.

pkg/tfbridge/schema.go
```diff
@@ -798,3 +794,49 @@ func CleanTerraformSchema(tfs map[string]*schema.Schema) map[string]*schema.Sche
...
 }
+
+// CoerceTerraformStringOutput coerces a field with the given Terraform schema to a PropertyValue. Terraform often
+// produces Go values of type String for resources whose schema indicates that the type is not String - usually Int,
+// Bool, or Float. In this case, CoerceTerraformStringOutput inspects the schema and, if the schema is present, coerces
+// the string value to the requested type and returns a PropertyValue for it.
+//
+// If the schema is not present (i.e. this is a Pulumi-specific field), there is an error coercing the string value,
+// or the schema indicates the field is not of type Int, Bool, or Float, the string is returned verbatim.
+func CoerceTerraformStringOutput(sch *schema.Schema, stringValue string) resource.PropertyValue {
```
pgavlin: How about something like

```suggestion
func CoerceTerraformString(to schema.ValueType, s string) interface{} {
    switch sch.Type {
    case schema.TypeInt:
        ...
    case schema.TypeBool:
        ...
    case schema.TypeFloat:
        ...
    default:
        return s
    }
}
```

This will make it easier to consume from `tf2pulumi` (or other libraries that don't deal in `resource.PropertyValue`).

swgillespie: sounds good, will do.

pkg/tfbridge/schema.go
```diff
@@ -480,8 +484,20 @@ func MakeTerraformOutput(v interface{},
...
+		}
+
+		// Otherwise, it might be a string that needs to be coerced to match the Terraform schema type. Coerce the
+		// string to the Go value of the correct type and, if the coercion produced something different than the string
+		// value we already have, re-make the output.
+		coerced, err := CoerceTerraformString(tfs.Type, t)
+		if err != nil || coerced == t {
+			return resource.NewStringProperty(t)
+		}
+		return MakeTerraformOutput(coerced, tfs, ps, assets, rawNames)
```
pgavlin: FWIW, I think this could have just been `resource.NewPropertyValue(coerced)`

swgillespie: oh, nice - I did not know that existed.

Issue #331, opened 2019-02-19
TFGen needs to accept InvokeOptions on all data sources


swgillespie:
Original report by @rswail:

> Each of the AWS Resource classes (eg Vpc, Instance) take an opts parameter that can include an alternate provider. This allows resources to be created and tracked across regions.

> However the equivalent get_x functions (eg get_vpc) do not, which means that only the default provider will be called when the function is invoked in pulumi.do_invoke.

> This doesn't allow cross region queries via alternate providers.

TFGen needs to emit an `opts` argument on each data source that allows for an instance of [InvokeOptions](https://pulumi.io/reference/pkg/python/pulumi/index.html#pulumi.InvokeOptions) to be passed.

PR #334, opened 2019-02-20
Allow passing of an InvokeOptions instance to Invoke

swgillespie:
Data sources implemented using Invoke can accept an InvokeOptions
parameter to control the invocation, including the provider to use to
perform the invocation. This commit exposes InvokeOptions to
tfgen-generated packages which previously did not have codegen for this
feature.

Fixes https://github.com/pulumi/pulumi-terraform/issues/331.

### DIFF
--- a/pkg/tfgen/generate_python.go
+++ b/pkg/tfgen/generate_python.go
@@ -390,7 +390,7 @@ func (g *pythonGenerator) emitRawDocComment(w *tools.GenWriter, comment, prefix 
 
 func (g *pythonGenerator) emitPlainOldType(w *tools.GenWriter, pot *plainOldType) {
 	// Produce a class definition with optional """ comment.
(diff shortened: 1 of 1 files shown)


pkg/tfgen/generate_python.go
```diff
@@ -390,7 +390,7 @@ func (g *pythonGenerator) emitRawDocComment(w *tools.GenWriter, comment, prefix
 
 func (g *pythonGenerator) emitPlainOldType(w *tools.GenWriter, pot *plainOldType) {
 	// Produce a class definition with optional """ comment.
-	w.Writefmtln("class %s(object):", pyClassName(pot.name))
+	w.Writefmtln("class %s:", pyClassName(pot.name))
```
swgillespie: This was a drive-by change, deriving from `object` explicitly isn't good Python style (Previously it was required in Python 2 for esoteric reasons, Python 3 considers it to be implicit).

Issue #327, opened 2019-02-13
Emitted __all__ member references packages that do not exist


swgillespie:
In the `gcp` package:

```python
__all__ = ['appengine', 'bigquery', 'bigtable', 'billing', 'binaryauthorization', 'cloudbuild', 'cloudfunctions', 'composer', 'compute', 'config', 'container', 'containeranalysis', 'dataflow', 'dataproc', 'dns', 'endpoints', 'filestore', 'folder', 'iam', 'kms', 'logging', 'monitoring', 'organizations', 'projects', 'pubsub', 'redis', 'resourcemanager', 'runtimeconfig', 'serviceAccount', 'sourcerepo', 'spanner', 'sql', 'storage'
```

The `serviceAccount` package does not exist - it should instead be `service_account`, the snake_cased name.

PR #335, opened 2019-02-20
Use Python name to refer to submodules in __all__

swgillespie:
Prior to this commit we were using a module's Pulumi name in __all__,
which is supposed to list the full set of submodules contained within a
module. Unfortunately this did not correspond to the on-disk module
format, which is a Python name (camel-cased), and thus it was impossible
to import this module.

This commit fixes the issue by using the Python name to refer to
submodules when generating __all__.

Fixes https://github.com/pulumi/pulumi-terraform/issues/327.


swgillespie:
> Do we need to push this out to all of the providers with submodules?

We do eventually, but this issue only affects the GCP provider as far as I can tell. I think this issue is pretty minor and can hold off until we do the next major rollup push.

PR #339, opened 2019-02-22
Update README.md

dustinfarris:
I was fumbling through here trying to figure out how to port a new repo, but looks like there is a separate repo that makes that more straightforward that I probably should have started with.


jen20:
Thanks @dustinfarris!

PR #345, opened 2019-02-28
Add an HCL fixer-upper

pgavlin:
These changes add a helper that can fix up certain types of syntactic
errors in HCL so that it can be parsed and converted by tf2pulumi. In
particular, the helper is able to fix the following errors:

- Misisng values in object properties
- Missing values in lists

It will also remove illegal tokens that follow a missing value.

### DIFF
--- a/pkg/tfgen/docs_hcl.go
+++ b/pkg/tfgen/docs_hcl.go
@@ -1,0 +1,289 @@
+// Copyright 2016-2018, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tfgen
+
+import (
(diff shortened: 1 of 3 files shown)


pgavlin:
This is the fix for https://github.com/pulumi/tf2pulumi/issues/73.

We could probably get _even more clever_ here and patch missing values using a special interpolation that `tf2pulumi` understands so that it could emit correctly-typed replacements (or elide the value itself).

pgavlin:
Diffs are clean, and the examples we lost due to the removed CIDR blocks are back: 
https://github.com/pulumi-bot/pulumi-aws/compare/469ac54697a39955a9e795aa93c7b7b1b0473dc2...3097832023b10773f24ea36217e4ed14a319caf4

pkg/tfgen/docs_hcl.go
```diff
@@ -0,0 +1,289 @@
...
+		t := h.scanner.Scan()
+
+		// Chew through comments. We save the offset of the first comment token and stamp it onto the first non-comment
+		// token we see so that patches (if any) will occur before comments. For example, given the following:
+		//
+		//     resource "aws_s3_bucket" "foo" {
+		//         name = # put a name here
+		//     }
+		//
+		// If we patch the missing property value, we want the patch to occur before "# put a name here". If we simply
```
swgillespie: Does this imply that we're dropping comments in the output stream? So your example would end up fixing to

```hcl
resource "aws_s3_bucket" "foo" {
    name = ""
}
```

pgavlin: Good question! It does not. The comments are effectively treated as part of the token they precede, and will be copied to the output with that token.

The only time we may drop comments is if they precede an `ILLEGAL` token that gets eluded by `patch`.

PR #346, opened 2019-03-01
Elide autoname properties in examples

pgavlin:
Use the new autoname elision in `tf2pulumi` to elide auto-named
properties in examples.


pgavlin:
This is a companion change to https://github.com/pulumi/tf2pulumi/pull/79.

https://github.com/pulumi/tf2pulumi/pull/79 actually depends on this PR, as it required [this bugfix](https://github.com/pulumi/pulumi-terraform/pull/346/files#diff-a2c3fec424d5eefcb8f9fa115d32b07fR347). The circular dependency won't matter in practice, however, so long as these changes are not rolled out to provider repos until `tf2pulumi` has been updated.

pgavlin:
@jen20 it is not: I tried to leave the comment above earlier, but apparently it didn't take ಠ_ಠ

PR #347, opened 2019-03-07
Adopt changes to Read and Diff

pgavlin:
1. Return resource inputs as well as resource state from Read()
2. Return a list of properties that changed from Diff()

The approach taken for (1) is very straightforward: any property that we
recognize as an input (i.e. it is either Required or Optional) is copied
to the input bag. Strictly speaking, this is an overestimate of the
necessary input set: some Optional properties may also be Computed, in
which case they may have been set by the provider itself and thus would
not be part of the user-specified input set.

(2) is very straightforward: the changed property names are simply
collected while we are looking for RequiresNew diffs. The names of the
returned properties have been fixed so that they only refer to top-level
properties and are properly converted to Pulumi names. This is necessary
in order for the engine (and the user) to make sense of the names.

The engine takes advantage of (2) to provide a good experience when (1)
overestimates the input set. Normally, such an overestimate could result
in a large number of spurious diffs presented to the user. Generally
speaking, though, TF considers the removal of an Optional Computed
property to be a no-diff change; filtering the diffs that are displayed
to only those indicated by the provider removes these spurious diffs.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -873,6 +873,7 @@
     "github.com/stretchr/testify/assert",
     "golang.org/x/net/context",
     "google.golang.org/grpc/codes",
+    "google.golang.org/grpc/status",
   ]
   solver-name = "gps-cdcl"
   solver-version = 1
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -161,6 +161,14 @@
   version = "v1.32.0"
 
 [[projects]]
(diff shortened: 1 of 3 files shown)


pgavlin:
The codegen tests are failing due to some trickiness in hack-vendor around the update to `github.com/pulumi/pulumi`. These changes do not affect codegen.

PR #348, opened 2019-03-08
Allow ResourcePrefix to be different to Name

jen20:
Some Terraform providers have a different name to the prefix on all of the resources - for example the Google Beta provider is named `google-beta`, but all of the resources have the prefix `google`. Previously, this caused documentation generation to fail - we can resolve that by introducing a separate ResourcePrefix member on ProviderInfo, and using that if it is set.

Marked as P1 since it blocks the Google Beta provider which is also P1.


jen20:
The downstream failure here is because of a mismatched pulumi/pulumi-terraform version in each downstream provider.

pgavlin:
This will probably confuse `tf2pulumi` as well... does that seem to be the case?

jen20:
I think `tf2pulumi` should be ok, but will test it with a provider using this once we have one.

PR #349, opened 2019-03-09
Track default population in inputs

pgavlin:
Add a reserved property to each map value present in an input bag that
tracks which keys in the map were populated using default values. When
performing default application, only use a value from an old input bag
if it was itself a default value.

This allows the following sequence of events:
- Create a resource with a value explicitly specified for property P
- Remove the value for property P and update the resource. The new value
  for property P will be whatever its default value is, if any.

Fixes https://github.com/pulumi/pulumi-aws/issues/439.

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -34,8 +34,11 @@ import (
 	"github.com/pulumi/pulumi/pkg/util/contract"
 )
 
-// defaultsKey is the name of the property that is used in inputs and state to track which input property
-// values were populated with default values.
+// defaultsKey is the name of the input property that is used to track which property keys were populated using
+// default values from the resource's schema. This information is used to inform which input properties should be
+// populated using old defaults in subsequent updates. When populating the default value for an input property, the
+// property's old value will only be used as the default if the property's key is present in the defaults list for
+// the old property bag.
 const defaultsKey = "__defaults"
 
 // AssetTable is used to record which properties in a call to MakeTerraformInputs were assets so that they can be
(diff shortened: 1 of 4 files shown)


pkg/tfbridge/schema.go
```diff
@@ -843,7 +883,42 @@ func CoerceTerraformString(schType schema.ValueType, stringValue string) (interf
 	return stringValue, nil
 }
 
-func extractInputsFromOutputs(urn resource.URN, outs resource.PropertyMap,
+func propagateDefaultAnnotations(oldInput, newInput resource.PropertyValue) {
```
pgavlin: This code is latent, and will become active once https://github.com/pulumi/pulumi/pull/2531 and an as-of-yet-unpublished PR land.

In essence: once we begin refreshing inputs, doing so will destroy the `__defaults` annotations this PR adds. In order to avoid that, we need to hand `Read` the current inputs as well as the current state. We can then propagate the `__defaults` annotations for properties whose values have not changed.

pgavlin:
Ping.

pgavlin:
Code-gen tests will fail here b/c of the update to `pulumi/pulumi`. These changes do not affect codegen.

pgavlin:
A demonstration of the effects of this change may prove useful. Below is a before/after of a stack that creates a versioned S3 bucket and then attempts to disable versioning. In the before case, `pulumi-terraform` picks up the old value of `true` and reuses it when the property is removed, so no diff is present and no update occurs. In the after case, `pulumi-terraform` is able to detect that the old value for `enabled` was not a default and should not be used as one, so a diff is present and the update occurs.

Before: https://asciinema.org/a/Drf7Mx4QLRd4fIZeL6UjbgKLY
After: https://asciinema.org/a/VbJSbHU3775nSoaxwqRnYSyGq

PR #350, opened 2019-03-09
Insert _ between words and numbers during name mangling

jen20:
Fixes pulumi/pulumi-azure#157.

When applied to the Azure provider, this allows the following previously failing program to run to completion:

```typescript
import * as azure from "@pulumi/azure";

// Create an Azure Resource Group
const resourceGroup = new azure.core.ResourceGroup("resourceGroup", {
    location: "West Europe",
});

const appServicePlan = new azure.appservice.Plan("test", {
    location: resourceGroup.location,
    resourceGroupName: resourceGroup.name,

    sku: {
        tier: "Standard",
        size: "S1"
    }
});

const appService = new azure.appservice.AppService("test", {
    location: resourceGroup.location,
    resourceGroupName: resourceGroup.name,
    appServicePlanId: appServicePlan.id,

    siteConfig: {
        dotnetFrameworkVersion: "v4.0",
        use32BitWorkerProcess: true,
        scmType: "LocalGit",
    },

    appSettings: {
        "SOMEKEY": "SomeValue",
    }
});
```

One thing that is unclear is whether this is correct in every case (i.e. whether Terraform providers are consistent with using underscores between words suffixed with numbers).

### DIFF
--- a/pkg/tfbridge/names.go
+++ b/pkg/tfbridge/names.go
@@ -34,6 +34,13 @@ func PulumiToTerraformName(name string, tfs map[string]*schema.Schema) string {
 				result += "_"
 			}
 			result += string(unicode.ToLower(c))
+		} else if c >= '0' && c <= '9' {
+			// If the last character is not a number, add a separator
+			lastChar := result[len(result)-1]
+			if lastChar < '0' || lastChar > '9' {
+				result += "_"
+			}
+			result += string(c)
 		} else {
 			result += string(c)
 		}
--- a/pkg/tfbridge/names_test.go
+++ b/pkg/tfbridge/names_test.go
@@ -29,6 +29,7 @@ func TestPulumiToTerraformName(t *testing.T) {
 	assert.Equal(t, "test_name", PulumiToTerraformName("test_name", nil))
 	assert.Equal(t, "test_name_", PulumiToTerraformName("testName_", nil))
 	assert.Equal(t, "t_e_s_t_n_a_m_e", PulumiToTerraformName("TESTNAME", nil))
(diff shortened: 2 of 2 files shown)


jen20:
Hmm, some of the other tests suggest that this is not going to work since we may rely on mangling the previous way, it needs further investigation into the various providers to see what we actually should do here.

lukehoban:
I do think there is a significant risk that the additional `_` will trigger issues with resources that do not allow that character. These are also exceedingly hard to track down a-priori. 

It’s not clear to me that the value of this is worth the disruption. If we do decide to take the disruption now or soon, I expect we’ll want to go further and move a more configurable autonamr mechanism like the one used in pulumi-azure into the core Pulumi-Terraform so that the autonaming can be tweaked as needed if/when the new defaults lead to issues.

joeduffy:
Do we know how many properties suffer from this problem? It may be worth a manual audit with some overrides in the schema mapping (offhand, I can't recall if we permit such renaming?)

lukehoban:
Retracted my previous comment as it was about an unrelated issue.

> Do we know how many properties suffer from this problem? 

I think it must be very few, as this is as far as I know the only property we've ever seen this reported for.  There are many properties which don't behave this way in canonical Terraform naming - like `http2_enabled` (not `http_2_enabled`) and `ipv6_address_count` (not `ipv_6_address_count`).

I assume the right thing to do here is just manually override the `name` of this property in `pulumi-azure`.

jen20:
I tried manually overriding `name` before making this change, but that did not work as expected - I assume because of a nuance with nested schema that I need to investigate to fix this correctly. Since Terraform doesn't appear to have a universally-accepted standard for mapping numbers like this, and changing it will likely break other things, I'll close this PR in favour of correcting the issue locally.

PR #351, opened 2019-03-12
Use the Pulumi property name to check old defaults

pgavlin:
The code as written records the Pulumi property name in the list of old
defaults but uses the TF property name for lookup. This does not work in
cases where the two do not match.


jen20:
I pull this downstream:

- https://github.com/pulumi/pulumi-azure/pull/202
and
- https://github.com/pulumi/pulumi-aws/pull/500

Neither saw the kinds of failure that were seen prior to this.

PR #354, opened 2019-03-18
Project TypeInt as floats in python

swgillespie:
Unlike JavaScript, Python clearly distinguishes between floats and
integers in their basic type hierarchy. Because the actual type of
values returned from tfbridge is always float, Python was observing that
fields whose schema type was TypeInt sometimes had float values, which
caused crashes and generally caused data sources to not behave well.

This fix causes tfgen to type all TypeInt fields as float, which has no
effect other than to change the type tests emitted by the return types
of data sources, which previously asserted that fields of type TypeInt
were literally of type int.


swgillespie:
Fixes this issue: https://github.com/pulumi/examples/pull/236#issuecomment-474033773

PR #360, opened 2019-04-02
Use Go Modules for dependency tracking

jen20:
This commit switches from dep to Go 1.12 modules for tracking dependencies of pulumi-terraform. We also upgrade Travis to Go 1.12.

Please do not merge this at this point!


jen20:
I've rebased this onto the latest master, and pointed our dependency on `pulumi/pulumi` back at `master` now that is tracked using Go modules.

PR #364, opened 2019-04-16
Add support for custom package name for nodejs

fernandocarletti:
I'm not very familiar with the Pulumi's codebase, please, let me know if there's a better approach for this.

Currently, I'm manually changing the package name whenever I build the package, this would allow me to set a custom name to it.

### DIFF
--- a/pkg/tfgen/generate_nodejs.go
+++ b/pkg/tfgen/generate_nodejs.go
@@ -732,9 +732,14 @@ func (g *nodeJSGenerator) emitNPMPackageMetadata(pack *pkg) error {
 	}
 	defer contract.IgnoreClose(w)
 
+	packageName := g.info.JavaScript.PackageName
+	if packageName == "" {
+		packageName = fmt.Sprintf("@pulumi/%s", pack.name)
+	}
+
 	// Create info that will get serialized into an NPM package.json.
 	npminfo := npmPackage{
-		Name:        fmt.Sprintf("@pulumi/%s", pack.name),
+		Name:        packageName,
 		Version:     "${VERSION}",
 		Description: g.info.Description,
 		Keywords:    g.info.Keywords,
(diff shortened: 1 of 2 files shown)


pkg/tfgen/generate_nodejs.go
```diff
@@ -732,9 +732,15 @@ func (g *nodeJSGenerator) emitNPMPackageMetadata(pack *pkg) error {
 	}
 	defer contract.IgnoreClose(w)
 
+	var packageName = g.info.JavaScript.PackageName
```
ellismg: NIT: It would be a little more idiomatic to write this as `packageName := g.info.JavaScript.PackageName`

pkg/tfgen/generate_nodejs.go
```diff
@@ -732,9 +732,15 @@ func (g *nodeJSGenerator) emitNPMPackageMetadata(pack *pkg) error {
...
+
```
ellismg: Maybe delete the blank line here?

ellismg:
Looks like a reasonable way to solve the problem!

One thing to note, right now, the CLI doesn't have great support for pulling plugins from external sources. It assumes that every plugin will be hosted on `api.pulumi.com`, so for folks that want to release new tf-bridged providers to large audiences, we either need to end up taking ownership or you need to deliver the provider out of band somehow, so it correctly ends up in a user's `~/.pulumi/plugins` folder.

fernandocarletti:
@ellismg We are creating a `mysql` provider using the https://github.com/pulumi/pulumi-tf-provider-boilerplate. I talked to @clstokes about it before, I have no issues on you guys taking ownership of the plugin. If that is the case, I would be glad to contribute to it.

I'm about to push the provider to https://github.com/LinioIT/, but if it is better to have it hosted with you, there's no problem, since we would have it forked in our side anyways.

lukehoban:
@ellismg Should we merge this as-is?  We now have support for hosting 3rd party plugins - so this is probably part of doing https://github.com/pulumi/pulumi-tf-provider-boilerplate/issues/13.

jen20:
This is now related to [`pulumi-terraform-bridge`][bridge], so I'll go ahead and close this PR out here. We'll work out how to apply it on the new repo shortly. Thanks @fernandocarletti!

[bridge]: https://github.com/pulumi/pulumi-terraform-bridge

PR #365, opened 2019-04-22
Strip "Open in Cloud Shell" buttons

justinvp:


justinvp:
The Travis "Test-downstream" jobs are failing because our `hack-vendor` script relies on `Gopkg.toml` and hasn't been updated for Go modules.

https://github.com/pulumi/scripts/blob/e6de0dd910e1158d7c20d6c2bc73b6ffc35103dd/hack-vendor/hack-vendor#L14-L16

Should we just disable those jobs until they've been updated to work with Go modules?

jen20:
The actual change looks good. I’ll put some effort into fixing the downstream checks tomorrow.

PR #366, opened 2019-04-26
Add support for "extra" config and config defaults

joeduffy:
This change adds two features to the bridge:

1) An ability to use config variables as default values for resource
   properties.

2) An ability to add "extra" config variables, which are Pulumi-only
   config variables added to tfgen packages. This can be used in
   conjunction with the above feature to plug in defaults for resource
   properties just solely based on these variables.

With these combined, we can achieve some simplifications to the Azure
resource provider, including a resource group's location coming from
config, plus all resources referring to that resource group inheriting
its location value.


joeduffy:
This is what was behind today's Hackathon demo.

If we decide to pursue this, I'll write a few tests before merging.

jen20:
This looks particularly useful! I think it may also be applicable in other providers.

jen20:
I've rebased this to pick up the downstream generation checks that were fixed recently.

PR #367, opened 2019-04-26
Enable stricter checks in our provider libs

CyrusNajmabadi:


pkg/tfgen/generate_nodejs.go
```diff
@@ -551,7 +551,7 @@ func (g *nodeJSGenerator) emitResourceType(mod *module, res *resourceType) (stri
 		w.Writefmtln("        let inputs: pulumi.Inputs = {};")
 		// The lookup case:
 		w.Writefmtln("        if (opts && opts.id) {")
-		w.Writefmtln("            const state: %[1]s = argsOrState as %[1]s | undefined;", stateType)
+		w.Writefmtln("            const state = argsOrState as %[1]s | undefined;", stateType)
```
CyrusNajmabadi: this was failing typechecking because we effectively were doing `const state: string = blah as string | undefined`.  It's now just `const state = blah as string | undefined`.

pkg/tfgen/generate_nodejs.go
```diff
@@ -811,7 +811,12 @@ func (g *nodeJSGenerator) emitTypeScriptProjectFile(pack *pkg, files []string) e
         "module": "commonjs",
         "moduleResolution": "node",
         "declaration": true,
-        "sourceMap": true
+        "sourceMap": true,
+        "stripInternal": true,
+        "experimentalDecorators": true,
+        "noFallthroughCasesInSwitch": true,
+        "forceConsistentCasingInFileNames": true,
+        "strict": true
```
CyrusNajmabadi: `strict: true` enables the gamut of strict tests.  I want this so that i can have my mixin code strictly analyzed

jen20:
I've rebased this to re-enable the downstream generation checks.

CyrusNajmabadi:
Thanks!

pkg/tfgen/generate_nodejs.go
```diff
@@ -507,7 +507,7 @@ func (g *nodeJSGenerator) emitResourceType(mod *module, res *resourceType) (stri
 			outcomment = "/*out*/ "
 		}
 
-		w.Writefmtln("    public %sreadonly %s: pulumi.Output<%s>;",
+		w.Writefmtln("    public %sreadonly %s!: pulumi.Output<%s>;",
```
CyrusNajmabadi: necessary since we don't actually set the values directly in our constructors (its done using reflection in registerResource).  this tells TS that that's ok and the values will be set through some mechanism.

PR #368, opened 2019-04-29
Use GetResourcePrefix vs string splitting

jen20:
In 0380ba21, we introduced the concept of a ResourcePrefix which differs from a Terraform Provider's name (to handle Google's Beta provider). We can use the `GetResourcePrefix` method on `ProviderInfo` to obtain this, instead of string splitting where the prefix is required.

### DIFF
--- a/scripts/test-downstream-impact
+++ b/scripts/test-downstream-impact
@@ -15,11 +15,12 @@ REPO_NAME="$1"
 
 git clone "https://github.com/pulumi/${REPO_NAME}.git" "$(go env GOPATH)/src/github.com/pulumi/${REPO_NAME}"
 
-# Run hack-vendor to update Terraform
+# Replace modle for Pulumi Terraform
 cd "$(go env GOPATH)/src/github.com/pulumi/${REPO_NAME}"
 
 git checkout -b "integration/pulumi-terraform/${TRAVIS_JOB_NUMBER}"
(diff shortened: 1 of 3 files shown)


jen20:
No, you're right - fixed. I've also modified this PR to fix up downstream generation tests again following the move to Go modules.

jen20:
@lukehoban I've added some tests around `camelPascalPulumiName` also.

scripts/test-downstream-impact
```diff
@@ -15,11 +15,12 @@ REPO_NAME="$1"
 
 git clone "https://github.com/pulumi/${REPO_NAME}.git" "$(go env GOPATH)/src/github.com/pulumi/${REPO_NAME}"
 
-# Run hack-vendor to update Terraform
+# Replace modle for Pulumi Terraform
 cd "$(go env GOPATH)/src/github.com/pulumi/${REPO_NAME}"
 
 git checkout -b "integration/pulumi-terraform/${TRAVIS_JOB_NUMBER}"
-"$(go env GOPATH)"/src/github.com/pulumi/scripts/hack-vendor/hack-vendor github.com/pulumi/pulumi-terraform
+GO111MODULE=on go mod edit -replace=github.com/pulumi/pulumi-terraform=../pulumi-terraform/
```
ellismg: Slick!

PR #369, opened 2019-05-01
Pass provider version to engine when invoking or registering resources

swgillespie:
Final part of https://github.com/pulumi/pulumi/issues/2389 for TF-based providers.

This PR augments the Python and Node code generators to explicitly pass the current package's version number as an argument to `Invoke`, `ReadResource`, and `RegisterResource`. We managed to accomplish querying our own version without using sed by:

1. For Node, requiring our own `package.json` and inspecting its version and
2. For Python, using `pkg_resources` (AFAIK part of a normal Python installation) to query the version of ourselves that's installed

In either case, if a version wasn't specified by the user, the version of the current package is sent to the engine. I've tested this end-to-end with the Random provider and it works like a charm (along with https://github.com/pulumi/pulumi/pull/2691).

### DIFF
--- a/pkg/tfgen/generate_nodejs_utilities.go
+++ b/pkg/tfgen/generate_nodejs_utilities.go
@@ -51,6 +51,12 @@ export function requireWithDefault<T>(req: () => T, def: T | undefined): T {
 }
 
 export function getVersion(): string {
-    return require('./package.json').version.slice(1);
+    let version = require('./package.json').version;
+    // Node allows for the version to be prefixed by a "v", while semver doesn't.
+    // If there is a v, strip it off.
+    if (version.indexOf('v') === 0) {
+        version = version.slice(1);
+    }
+    return version;
 }
 `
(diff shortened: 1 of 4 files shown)


pkg/tfgen/generate_nodejs_utilities.go
```diff
@@ -49,4 +49,8 @@ export function requireWithDefault<T>(req: () => T, def: T | undefined): T {
     }
     return def;
 }
+
+export function getVersion(): string {
+    return require('./package.json').version.slice(1);
```
ellismg: I think the intention of this slice is to remove the leading `v` from the version, correct?

`package.json` does allow for you to elide the leading `v`, and while all our generated packages will have this as an artifact of how we do version tagging, do you think we could augment this to not to the trim if the v isn't present?

ellismg:
Can you add a tracking item to the general golang issue about how we'll want to implement this for golang as well (I assume the implementation is just using the `pkg/version.Version` const).

pkg/tfgen/generate_nodejs_utilities.go
swgillespie: > package.json does allow for you to elide the leading v

Ah, I did not know that - I think you're right, we should do that.

swgillespie:
> (I assume the implementation is just using the pkg/version.Version const).

It would involve some `sed` on that file, but yeah, pretty much.

swgillespie:
@ellismg does this look OK to merge?

PR #373, opened 2019-05-09
Add ability to remove array elements which are empty objects

jen20:
This PR is potential way to fix the issue described in pulumi/pulumi-azure#2511, and several other variants of it we have seen in the past. It's unclear that removing empty array elements is always the right option, so we make it opt-in for resources where it is known to be effective instead of applying it generally.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -83,16 +83,38 @@ type DataSourceInfo struct {
 
 // SchemaInfo contains optional name transformations to apply.
 type SchemaInfo struct {
-	Name        string                 // a name to override the default; "" uses the default.
-	Type        tokens.Type            // a type to override the default; "" uses the default.
-	AltTypes    []tokens.Type          // alternative types that can be used instead of the override.
-	Transform   Transformer            // an optional idemponent transformation, applied before passing to TF.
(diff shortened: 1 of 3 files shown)


pkg/tfbridge/info.go
```diff
@@ -83,16 +83,38 @@ type DataSourceInfo struct {
...
-	Type        tokens.Type            // a type to override the default; "" uses the default.
-	AltTypes    []tokens.Type          // alternative types that can be used instead of the override.
-	Transform   Transformer            // an optional idemponent transformation, applied before passing to TF.
-	Elem        *SchemaInfo            // a schema override for elements for arrays, maps, and sets.
-	Fields      map[string]*SchemaInfo // a map of custom field names; if a type is missing, the default is used.
-	Asset       *AssetTranslation      // a map of asset translation information, if this is an asset.
-	Default     *DefaultInfo           // an optional default directive to be applied if a value is missing.
-	Stable      *bool                  // to override whether a property is stable or not.
-	MaxItemsOne *bool                  // to override whether this property should project as a scalar or array.
+	// a name to override the default; "" uses the default.
```
jen20: These changes are to appease the line length linter, as `gofmt` style makes several lines longer than the limit when the new value is added.

PR #374, opened 2019-05-11
Add @pulumi/terraform package and state.RemoteStateReference resource

jen20:
This PR adds a new NodeJS package to `pulumi/pulumi-terraform`, named `@pulumi/terraform`. The primary resource in the package is `terraform.state.RemoteStateReference`, which acts like a `pulumi.StackReference`, but for Terraform Remote State. This allows Pulumi programs to read values for infrastructure which was provisioned using Terraform.

The provider is implemented using the native RPC interface rather than the Terraform bridge, and the NodeJS SDK is handwritten. Since Python does not yet have "Read" support, only a NodeJS version is available.

Currently types only exist for two Terraform Remote State backends - S3 and Consul. The rest are straightforward to map, but I thought I'd get the bulk of this out for review before doing so, since it is quite time consuming.


jen20:
The `RemoteStateReference` resource is used as follows:

```typescript
import * as tf from "@pulumi/terraform";

const rs = new tf.state.RemoteStateReference("s3state", {
    backendConfig: {
        backendType: "s3",
        bucket: "pulumi-terraform-state-test",
        key: "test/terraform.tfstate",
        region: "us-west-2"
    }
});

export const vpcId = rs.getOutput("vpc_id");
export const publicSubnetIds = rs.getOutput("public_subnet_ids");
export const bucketArn = rs.getOutput("bucket_arn");
```

cmd/pulumi-resource-terraform/main.go
```diff
@@ -0,0 +1,10 @@
+package main
+
+import (
+	"github.com/pulumi/pulumi-terraform/pkg/provider"
+	"github.com/pulumi/pulumi-terraform/pkg/version"
+)
+
+func main() {
+	provider.Serve("derp", version.Version)
```
joeduffy: Should this be `"terraform"`? 😉

pkg/provider/cty.go
```diff
@@ -0,0 +1,109 @@
+package provider
```
joeduffy: FWIW, I wasn't sure what `cty` refers to.

joeduffy: Minor nit, missing the typical copyright headers.

sdk/nodejs/state/remoteStateReference.ts
```diff
@@ -0,0 +1,59 @@
...
+import {ConsulBackendConfig} from "./consulBackendConfig";
+
+/**
+ * The set of arguments for constructing a RemoteStateReference resource.
+ */
+export interface RemoteStateReferenceArgs {
+    /**
+     * The configuration of the Terraform Remote State backend, including type.
+     */
+    readonly backendConfig: S3BackendConfig | ConsulBackendConfig;
```
joeduffy: Since this is the key part and `workspace` is backend-specific already, should we consider just doing

```typescript
export type RemoteStateReferenceArgs = S3BackendConfig | ConsulBackendConfig;
```

and perhaps rename `S3BackendConfig` to `S3RemoteStateReferenceArgs`, etc? It'd slightly simplify the use of the type. (Possibly a bad idea ... just asking ...)

cmd/pulumi-resource-terraform/main.go
jen20: Yes! Somewhat surprised that still worked...

sdk/nodejs/state/remoteStateReference.ts
jen20: We could do this - in fact it was the original design. The reason I decided against it in the end was needing to delete unused keys from the config before it is passed to the backend - it seems simpler to only have to delete the discriminator key than anything in the config.

pkg/provider/cty.go
jen20: This should be fixed across all files now.

jen20: I've added a comment clarifying that `go-cty` is the configuration format used by remote state backends in Terraform.

pkg/provider/remote_state.go
```diff
@@ -0,0 +1,140 @@
...
+	be "github.com/hashicorp/terraform/backend"
+	backendInit "github.com/hashicorp/terraform/backend/init"
+	"github.com/pulumi/pulumi/sdk/proto/go"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func remoteStateReferenceRead(ctx context.Context, req *pulumirpc.ReadRequest) (*pulumirpc.ReadResponse, error) {
+	// Prevent Terraform from logging minutia
+	log.SetOutput(ioutil.Discard)
```
lukehoban: Curious - why do we want to do this?  Don't we want to report back logs for diagnostics puproses?

I know we have `--debug` which today gets passed through to providers and can be used to turn on verbose Terraform Provider logging - can we support this in the same way here?

sdk/nodejs/state/remoteStateReference.ts
lukehoban: > The reason I decided against it in the end was needing to delete unused keys from the config before it is passed to the backend

I'm not sure I follow why that would be necessary.  At the JavaScript level (the runtime object), the same subsets of properties are supported in either case - right?

jen20: In JavaScript yes - but ultimately we get a `structpb` in the providers handler, and have to manipulate that. We cannot call `PrepareConfig` or `CoerceValue` on the `go-cty` object representing the config struct with extraneous values in the map we pass in, so values have to be deleted. I've changed it to make everything a top level property however - there are only two fields to delete instead of one.

lukehoban: > with extraneous values in the map we pass in

How are extraneous values getting in here?  How does the TypeScript type impact what values are in here?  (all that should matter is what JavaScript object is passed in)

jen20: We get the properties in the shape they appear in the object here. If config was in an object, there is nothing to do:

```javascript
{
    config: {
        // all keys in here go to the go-cty object
    },
    workspace: "x" // this _must not_ go to the go-cty object
}
```

Whereas if we have all of the config keys in one object, we have to construct a new object with just the keys in the config schema before being able to pass it to the go-cty object.

However, since there is only one additional key in there, it's not a huge problem (and is now implemented).

pkg/provider/remote_state.go
jen20: Quite possibly - as it is now, the log comes back over the standard diagnostics and feeds back INFO level messages in all cases. I'll look at how we deal with this elsewhere.

jen20: Sorry, I mean _without_ discarding the log - not as it is now.

jen20:
I've also now added type definitions for all of the non-deprecated standard backends. I think some more work will be required for the "enhanced" backends for Terraform Enterprise and `local`.

pkg/provider/remote_state.go
lukehoban: See for example: https://github.com/pulumi/pulumi-terraform/blob/d089c9cf69cff4594f59d0e80c5afb7efa596fc8/pkg/tfbridge/provider.go#L196

lukehoban:
>  I think some more work will be required for the "enhanced" backends for Terraform Enterprise and local.

These feel like they will be fairly common.  When you say "more work" do you expect it is incremental, or very different?

sdk/nodejs/state/remoteStateReference.ts
lukehoban: I see what you are saying. 

> However, since there is only one additional key in there, it's not a huge problem (and is now implemented).

Yeah - this seems okay.

pkg/provider/remote_state.go
jen20: Yup, that's what we want.

jen20: OK this should be resolved now.

jen20:
All of the backends are now supported, and there is an example of the `local` backend running in CI.

PR #375, opened 2019-05-14
Don't emit `config.require` calls

lukehoban:
We read `config` to populate variables with the values of configuration.  This is useful, but not all usage of the provider depends on Pulumi configuration, as explicit providers (1st class providers) do not configure themselves with Pulumi config.
    
While we still want to populate the `config` module for a provider with the Pulumi configuration, we do not want to fail when it is missing - we just want to leave the variables undefined.  Therefore, we should not emit `require` calls.
    
Helpful error messages are still provided to the user, but they are provided through a more centralized contract with the provider Configure RPC instead of via ad-hoc `require` calls in the SDKs.
    
Fixes pulumi/pulumi#2725


lukehoban:
An example of the codegen changes this causes: https://github.com/pulumi/pulumi-aws/pull/558.

PR #381, opened 2019-05-22
Do not flow version information with the request

ellismg:
In absence of a fix for pulumi/pulumi#2753, passing a version here
means that every time you upgrade your package, all the resources it
managed will be replaced.

That's not an experience we want, so for now we are backing out this
part of the change.

### DIFF
--- a/pkg/tfgen/generate_nodejs.go
+++ b/pkg/tfgen/generate_nodejs.go
@@ -580,14 +580,15 @@ func (g *nodeJSGenerator) emitResourceType(mod *module, res *resourceType) (stri
 	}
 	w.Writefmtln("        }")
 
+	// TODO[pulumi/pulumi#2753]: We should re-enable this code once we have a solution for #2753.
 	// If the caller didn't request a specific version, supply one using the version of this library.
-	w.Writefmtln("        if (!opts) {")
-	w.Writefmtln("            opts = {}")
-	w.Writefmtln("        }")
-	w.Writefmtln("")
-	w.Writefmtln("        if (!opts.version) {")
-	w.Writefmtln("            opts.version = utilities.getVersion();")
-	w.Writefmtln("        }")
+	// w.Writefmtln("        if (!opts) {")
+	// w.Writefmtln("            opts = {}")
+	// w.Writefmtln("        }")
+	// w.Writefmtln("")
+	// w.Writefmtln("        if (!opts.version) {")
+	// w.Writefmtln("            opts.version = utilities.getVersion();")
+	// w.Writefmtln("        }")
 
 	// Now invoke the super constructor with the type, name, and a property map.
 	w.Writefmtln(`        super("%s", name, inputs, opts);`, res.info.Tok)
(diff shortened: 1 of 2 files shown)


swgillespie:
> In absence of a fix for pulumi/pulumi#2753, passing a version here
means that every time you upgrade your package, all the resources it
managed will be replaced.

Can we instead comment out the code where the engine deserializes a version? Here: https://github.com/pulumi/pulumi/blob/31bd4632641dd83d3fefd0fe2848764d81c7ed18/pkg/resource/deploy/source_eval.go#L497-L508

If the objective is to disable the behavior, we'll get the same result if we make this change in the CLI without having to roll this out to every package, which doesn't help the packages we've already shipped.

lukehoban:
> Can we instead comment out the code where the engine deserializes a version

I'm sympathetic to that (and we should possibly do that as well until this is resolved) - but doing only that would mean you have to be careful to not use certain combinations of provider and CLI versions.  I have a feeling it will be harder to shut down the spread of this issue if we only do that.  Thoughts?

swgillespie:
> I'm sympathetic to that (and we should possibly do that as well until this is resolved) - but doing only that would mean you have to be careful to not use certain combinations of provider and CLI versions.

How so? I think that doing this change does exactly what you're afraid of, in that it creates a sequence of packages that sent their version, stopped sending their version, and then at some point in the future will send their version again. If we make this change only in the CLI, we'd have a CLI that occasionally falls back to the previous behavior and lights-up to the correct, version-respecting behavior when we fix the issue.

swgillespie:
In my mind the problem is not "packages are sending their version", it's "the engine does the wrong thing with the version". I'd rather revert the second before the first, since the first thing is something we're going to have to do anyways.

swgillespie:
That said, either way I don't think it's possible to be sure which course of action is better here without a robust suite of tests to lean upon, which we don't have. I'm :+1: with this for putting out the fire for now.

ellismg:
To add to @swgillespie I think no matter what we do we are going to have a set of CLIs and packages that don't place nicely together.  Basically, if you are using a CLI newer than 0.17.7 (which understood that versions could come along with requests, and would include them in the provider name) and a new package that sends the version, you are going to have issues.

As Sean says, we can either stop sending versions of or we can stop interpreting them in the engine.  I do believe the later is better, vs trying to flow this change to every provider.  I think it also provides a much nicer story for folks that run into this issue (upgrade your version of the CLI) vs upgrade an arbitrary number of packages.

Note that: https://github.com/pulumi/pulumi-kubernetes/pull/569 would also introduce this problem for kubernetes.

ellismg:
I think that @swgillespie has convinced me that doing this in the CLI is the right first step, so I'm not going to do this here now and I'll open a PR in pulumi/pulumi to disable the version parsing.

swgillespie:
>  I think it also provides a much nicer story for folks that run into this issue (upgrade your version of the CLI) vs upgrade an arbitrary number of packages.

I think that in general, if we can fix an issue by taking a CLI fix, we should. Users are firmly in control of what version of the CLI they use, but they are not always in control of what version of a package they use, particularly in the case of node, where users are at the mercy of yarn and npm.

ellismg:
We've decided to do fixes here in both the CLI and in the the codegen for now.

PR #382, opened 2019-06-04
Generate `isInstance` type guard for every `CustomResource`

hausdorff:
`Resource.isInstance` is a static method that provides a type guard
allowing users to check whether an arbitrary object is a `Resource`.

This commit implements this function for all implementations of
`Resource`. So, e.g., `s3.Bucket.isInstance(foo)` is a type guard for
`Bucket`.

### DIFF
--- a/pkg/tfgen/generate_nodejs.go
+++ b/pkg/tfgen/generate_nodejs.go
@@ -479,6 +479,21 @@ func (g *nodeJSGenerator) emitResourceType(mod *module, res *resourceType) (stri
 		w.Writefmtln("")
 	}
 
+	w.Writefmtln("    /** @internal */")
+	w.Writefmtln("    public static readonly __pulumiType = '%s';", res.info.Tok)
+	w.Writefmtln("")
+	w.Writefmtln("    /**")
+	w.Writefmtln("     * Returns true if the given object is an instance of %s.  This is designed to work even", name)
+	w.Writefmtln("     * when multiple copies of the Pulumi SDK have been loaded into the same process.")
+	w.Writefmtln("     */")
+	w.Writefmtln("    public static isInstance(obj: any): obj is %s {", name)
+	w.Writefmtln("        if (obj === undefined || obj === null) {")
+	w.Writefmtln("            return false;")
+	w.Writefmtln("        }")
+	w.Writefmtln("        return obj['__pulumiType'] === %s.__pulumiType;", name)
+	w.Writefmtln("    }")
+	w.Writefmtln("")
+
 	// Emit all properties (using their output types).
 	// TODO[pulumi/pulumi#397]: represent sensitive types using a Secret<T> type.
 	ins := make(map[string]bool)
@@ -591,7 +606,7 @@ func (g *nodeJSGenerator) emitResourceType(mod *module, res *resourceType) (stri
 	// w.Writefmtln("        }")
 
 	// Now invoke the super constructor with the type, name, and a property map.
-	w.Writefmtln(`        super("%s", name, inputs, opts);`, res.info.Tok)
+	w.Writefmtln(`        super(%s.__pulumiType, name, inputs, opts);`, name)
 
 	// Finish the class.
 	w.Writefmtln("    }")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ This CHANGELOG details important changes made in each version of the
 
 ## v0.18.3 (Unreleased)
 
+### Improvements
+
+- Automatically generate `isInstance` type guards for implementations of `Resource`.
+
 ## v0.18.2 (Released May 28th, 2019)
 
 - Improved the package `README` file to reflect usage of the `@pulumi/terraform`


pkg/tfgen/generate_nodejs.go
```diff
@@ -479,6 +479,29 @@ func (g *nodeJSGenerator) emitResourceType(mod *module, res *resourceType) (stri
...
+		w.Writefmtln("     */")
+		w.Writefmtln("    public static isInstance(obj: any): obj is %s {", name)
+		w.Writefmtln("        if (obj === undefined || obj === null) {")
+		w.Writefmtln("            return false;")
+		w.Writefmtln("        }")
+		w.Writefmtln("")
+		w.Writefmtln("        const t = obj['__pulumiType'];")
+		w.Writefmtln("        if (typeof t !== 'string') {")
+		w.Writefmtln("            return false;")
+		w.Writefmtln("        }")
```
lukehoban: This `typeof` test seems unnecessary - it cannot be `===` below` if it's not the same type.

pkg/tfgen/generate_nodejs.go
```diff
@@ -479,6 +479,29 @@ func (g *nodeJSGenerator) emitResourceType(mod *module, res *resourceType) (stri
 		w.Writefmtln("")
 	}
 
+	w.Writefmtln("    private static readonly __pulumiType = '%s';", res.info.Tok)
```
lukehoban: Can we make this public?  That would possibly address https://github.com/pulumi/pulumi/issues/2754.`.

If we do make it pubic - I suspect it is just `type`?

cc also @CyrusNajmabadi for thoughts.

pkg/tfgen/generate_nodejs.go
```diff
@@ -479,6 +479,29 @@ func (g *nodeJSGenerator) emitResourceType(mod *module, res *resourceType) (stri
...
+
+	if !res.IsProvider() {
```
lukehoban: Why do we leave this out for providers?  It seems reasonable to want to write queries that use `isInstance` that look at provider instances and verify things about them.

CyrusNajmabadi:
> This commit implements this function for all implementations of
Resource. So, e.g., s3.Bucket.isInstance(foo) is a type guard for
Bucket.

Note: this actually sounds like it could be a breaking change.  i.e. someone might have written `Bucket.isInstance` before, having it only check that it was a resource.  It would previously pass, but now woudl fail.

We've been dealing with that by having specific things like `isLoadBalancer` instead of `isInstance` to make it very clear what you're asking and to make it so that static inheritance doesn't play a role here.

pkg/tfgen/generate_nodejs.go
CyrusNajmabadi: I think it would be ok to have this public.

pkg/tfgen/generate_nodejs.go
CyrusNajmabadi: 👍

pkg/tfgen/generate_nodejs.go
```diff
@@ -591,7 +614,7 @@ func (g *nodeJSGenerator) emitResourceType(mod *module, res *resourceType) (stri
 	// w.Writefmtln("        }")
 
 	// Now invoke the super constructor with the type, name, and a property map.
-	w.Writefmtln(`        super("%s", name, inputs, opts);`, res.info.Tok)
+	w.Writefmtln(`        super(%s.__pulumiType, name, inputs, opts);`, name)
```
CyrusNajmabadi: cute.

CyrusNajmabadi:
Overall i'm strongly in support of having these isXXX members.  My primary concern is the inheritance problem.  I think a better approach might be to generate members like `Bucket.isBucket`.  That way there are no collisions.  Thoughts @lukehoban @hausdorff ?

lukehoban:
> Note: this actually sounds like it could be a breaking change. i.e. someone might have written Bucket.isInstance before, having it only check that it was a resource. It would previously pass, but now woudl fail.

This is true, but seem sufficiently rare that I'm not deeply worried about the break.  If anything, it's likely more expected that `Bucket.isInstance` checks it it's a `Bucket` not just a `Resource`.

CyrusNajmabadi:
> This is true, but seem sufficiently rare that I'm not deeply worried about the break. If anything, it's likely more expected that Bucket.isInstance checks it it's a Bucket not just a Resource.

You're right.  And i'm acutally wrong and was misremembering something.  At some point a bug was almost introduced because someone did ComponentResource.isInstance (but got Resource.isInstance) instead.  So we explicitly overrode the method for Custom/ComponentResource to give the expected behavior.

Given that, we *def* should follow through and do it for the rest of the resources.  so i'm ok with `isIsntance` as the naming/hiding strategy.

pkg/tfgen/generate_nodejs.go
hausdorff: This is basically copied from the pulumi/pulumi implementation of `isInstance`

hausdorff:
I strongly believe that users who call `Bucket.isInstance` want to know if it's a `Bucket` or not. So I think we're all in agreement that this shoudl stay.

pkg/tfgen/generate_nodejs.go
CyrusNajmabadi: the existing isInstance does things differently (and can't change because it literally needs to always operate the same in SxS scenarios), so you def don't have to follow it :)

Looking at it, i can see i was too paranoid.  It doesn't need the `typeof` check.  The `===` check is sufficient against that precise value that it's looking for.

hausdorff:
@lukehoban @ellismg @CyrusNajmabadi I think this is ready to go. We should try to ship soon so that this makes it into the next release of all the providers

CyrusNajmabadi:
Yup yup.   No objections from me.  Just glad we discussed it and made the informed choice :)

CyrusNajmabadi:
Standard annoying note: we should do this for python too.  Sorry.

pkg/tfgen/generate_nodejs.go
```diff
@@ -479,6 +479,20 @@ func (g *nodeJSGenerator) emitResourceType(mod *module, res *resourceType) (stri
 		w.Writefmtln("")
 	}
 
+	w.Writefmtln("    public static readonly __pulumiType = '%s';", res.info.Tok)
```
CyrusNajmabadi: i would just call this "type".  Though there's a small concern of you colliding with a resource field of that name.  i don't like it being both public and `__`.  either remove the `__` or make private.

hausdorff:
We didn't implement in Python on purpose, because it doesn't have type guards. Is it actually useful in this case?

CyrusNajmabadi:
> We didn't implement in Python on purpose, because it doesn't have type guards. Is it actually useful in this case?

Oh... hrmm..  I guess that's fair.  I think python doesn't have the option of even loading multiple versions of a package (which is why we added this).  

--

Looking, it actually looks like our python code has an `isinstance` function which is used to test things.  but for that guy, you pass in the class you want to test again.  i.e. `isinstance(obj, Resource)`.  So python devs already ahve something they can use here. 

Thanks!  Just fix the naming of hte `__` member and i'm good.

CyrusNajmabadi:
(or make it private).

pkg/tfgen/generate_nodejs.go
hausdorff: We decided offline to make this `@internal` and `public`.

CyrusNajmabadi:
LGTM.  ThankS!

pkg/tfgen/generate_nodejs.go
lukehoban: @hausdorff FWIW - I don't think this feedback was addressed?

PR #388, opened 2019-06-06
Force UTF-8 when opening README.rst

ellismg:
I believe that this will address issues like: https://github.com/pulumi/pulumi-aws/issues/587.

I'm going to try to confirm this actually fixes the issue on my windows machine, assuming I can get it to believe it is in the problematic locale and can reproduce the issue.


jen20:
This looks good in principle, but it also appears to break all of the downstream code generation?

ellismg:
> but it also appears to break all of the downstream code generation?

Yeah, it turns out we are still using python2 in some places, which did not support `encodings` here.  I am working on moving everything to python3 and then we will be able to land this.

stack72:
Superseded via #494

PR #395, opened 2019-06-19
Use a schematized reader in MakeTerraformResult.

### LINKED ISSUE #394: TF attribute -> Pulumi property value conversion is incorrect in the face of map keys that contain '.'
e.g. given the following TF state:
```
                        "attributes": {
                            "tags.%": "1",
                            "tags.pulumi.io/foo/bar": "baz",
                        },
```

we will generate this property bag:
```
{
    "tags": {
        "pulumi": {
            "io/foo/bar": "baz"
        }
    }
```

when we should be generating this property bag:
```
    "tags": {
        "pulumi.io/foo/bar": "baz"
    }
```

This can cause spurious diffs.

pgavlin:
This prevents bugs where `flatmap.Expand` would misinterpret the
contents of a resource's state with respect to the shape described by
the resource's schema. The most critical misinterpretation involves
submaps with keys that contain "." characters: in this case, the key is
misinterpreted as describing another submap rather than a string ->
string entry. For example, given the following TF state:

```
"attributes": {
    "tags.%": "1",
    "tags.pulumi.io/foo/bar": "baz",
},
```

flatmap will expand this to:
```
{
    "tags": {
        "pulumi": {
            "io/foo/bar": "baz"
        }
    }
```

instead of the expected
```
    "tags": {
        "pulumi.io/foo/bar": "baz"
    }
```

Fixes #394.

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -461,13 +461,24 @@ const metaKey = "__meta"
 // MakeTerraformResult expands a Terraform state into an expanded Pulumi resource property map.  This respects
 // the property maps so that results end up with their correct Pulumi names when shipping back to the engine.
 func MakeTerraformResult(state *terraform.InstanceState,
-	tfs map[string]*schema.Schema, ps map[string]*SchemaInfo) resource.PropertyMap {
+	tfs map[string]*schema.Schema, ps map[string]*SchemaInfo) (resource.PropertyMap, error) {
 	var outs map[string]interface{}
 	if state != nil {
 		outs = make(map[string]interface{})
 		attrs := state.Attributes
+
+		reader := &schema.MapFieldReader{
+			Schema: tfs,
+			Map:    schema.BasicMapReader(attrs),
+		}
 		for _, key := range flatmap.Map(attrs).Keys() {
-			outs[key] = flatmap.Expand(attrs, key)
+			res, err := reader.ReadField([]string{key})
+			if err != nil {
+				return nil, err
+			}
+			if res.Value != nil {
+				outs[key] = res.Value
+			}
 		}
 	}
 	outMap := MakeTerraformOutputs(outs, tfs, ps, nil, false)
@@ -479,7 +490,7 @@ func MakeTerraformResult(state *terraform.InstanceState,
 		outMap[metaKey] = resource.NewStringProperty(string(metaJSON))
 	}
 
-	return outMap
+	return outMap, nil
 }
 
 // MakeTerraformOutputs takes an expanded Terraform property map and returns a Pulumi equivalent.  This respects
@@ -522,6 +533,11 @@ func MakeTerraformOutput(v interface{},
 
 	if v == nil {
 		return resource.NewNullProperty()
+	}
+
+	// Marshal sets as their list value.
+	if set, isset := v.(*schema.Set); isset {
+		v = set.List()
 	}
 
 	// We use reflection instead of a type switch so that we can support mapping values whose underlying type is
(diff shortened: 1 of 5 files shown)


pgavlin:
This fix is kind of terrifying. The other option I explored was writing our own version of `flatmap.Expand`, but I am pretty confident that that would only get things differently wrong. At the very least, this should result in behavior that is consistent with the vast majority of TF providers, which generally use (indirectly) schematized readers when examining property values.

pgavlin:
This should also fix https://github.com/pulumi/pulumi-eks/issues/152.

pgavlin:
This should also fix https://github.com/pulumi/pulumi-eks/issues/152.

pkg/tfbridge/schema.go
```diff
@@ -461,13 +461,24 @@ const metaKey = "__meta"
...
+			Schema: tfs,
+			Map:    schema.BasicMapReader(attrs),
+		}
 		for _, key := range flatmap.Map(attrs).Keys() {
-			outs[key] = flatmap.Expand(attrs, key)
+			res, err := reader.ReadField([]string{key})
+			if err != nil {
+				return nil, err
+			}
+			if res.Value != nil {
```
pgavlin: This will only happen for keys that are not present in the resource's schema. One notable field in this category is the `id` field, which we will no longer populate. I do not believe that this should be problematic--the only values we expose through the SDKs today are values that are present in the schema, and we communicate the value of the `id` field out of band--but I am open to the idea of backfilling properties that are present in the flatmap but not in the schema. Opinions?

pgavlin: (FWIW, it would be _very_ easy to do something as simple as retain the `id` field)

lukehoban: For data-sources, I recall we had issues with not populating `id` in outputs?  Will this impact that?

Am I correct that you believe these changes result in no user-visible change in behaviour?  If there is expected user-visibile change in behaviour - what is it exactly?

pgavlin: > For data-sources, I recall we had issues with not populating id in outputs? Will this impact that?

This will not impact that, no: we explicitly set the ID property in our implementation of `Invoke`.

> Am I correct that you believe these changes result in no user-visible change in behaviour? If there is expected user-visibile change in behaviour - what is it exactly?

I believe that it is exceedingly unlikely that these changes will cause any changes to user-visible behavior besides the following:
- The `id` property will no longer be present in resource states. This will show up during a `refresh` as a removal. If this is undesirable (and it may well be), it is easy enough to restore this specific property.
- Maps with keys that contain characters that are semantically meaningful to flatmap will be properly marshaled/unmarshaled (this is the bugfix)

It is _possible_ that other properties that are not present in the schema will be missing, but those are only exposed to users via the checkpoint itself (not via the SDKs).

lukehoban: >  If this is undesirable (and it may well be), it is easy enough to restore this specific property.

Yes - that sounds undesirable.  Let's restore that the `id` property.

pgavlin: Done.

pkg/tfbridge/schema.go
```diff
@@ -524,6 +538,11 @@ func MakeTerraformOutput(v interface{},
 		return resource.NewNullProperty()
 	}
 
+	// Marshal sets as their list value.
+	if set, isset := v.(*schema.Set); isset {
+		v = set.List()
+	}
```
lukehoban: Totally curious - how is this related?

pkg/tfbridge/schema.go
```diff
@@ -461,13 +461,27 @@ const metaKey = "__meta"
...
-			outs[key] = flatmap.Expand(attrs, key)
+			res, err := reader.ReadField([]string{key})
+			if err != nil {
+				return nil, err
+			}
+			if res.Value != nil {
+				outs[key] = res.Value
+			}
+		}
+		if _, ok := outs["id"]; !ok {
```
lukehoban: Perhaps a comment on why this is needed?

pkg/tfbridge/schema.go
pgavlin: The `MapFieldReader` can produce `*schema.Set` values. `flatmap.Expand` would have instead produced `[]interface{}` values for these properties.

pkg/tfbridge/schema.go
pgavlin: Sure.

PR #401, opened 2019-07-02
Return detailed diffs.

pgavlin:
These changes add support for converting a Terraform instance diff into
a Pulumi detailed diff and returning the result from `Diff`.

Both a Terraform instance diff and a Pulumi detailed diff are composed
of a list of (changed property path, change type) tuples. Despite this
similarity, converting the former to the latter is complicated due to
a number of subtle differences.

- Path elements in the TF instance diff must be mapped to their Pulumi
  property names. The ambiguity of TF diff paths is addressed by driving
  the mapping using the Pulumi property paths as input.
- A TF instance diff does not always indicate that an entire composite
  value was removed. Instead, it may indicate that each component
  property of that value was removed. The conversion leaves this
  representation intact (though the engine may display such diffs as
  complete removal of the composite value).
- A TF instance diff does not indicate that a property was added.
  Instead, property adds are represented by a diff with an old value of
  the empty string. This is indistinguishable from a change in which the
  old property was present and its value was the empty string. The
  conversion addresses this by looking for diffs to keys that were
  only present in the new config and treating such diffs as adds.
- TF set elements are represented in Pulumi property bags by array
  elements, but are represented in TF instance diffs by their hashcode.
  Because the mapping from a set element to its hashcode is one-way, we
  must drive the conversion by taking the Pulumi property value that
  represents the set element, calculating its hashcode, and checking to
  see if there is a diff at that hashcode.

These issues--and their solutions--are also documented in
`makePropertyDiff`.


pgavlin:
The PR build is failing due to downstream dependency issues in the Azure repo.

PR #409, opened 2019-07-12
Improve TF docs generation

ellismg:
- Don't drop URLs to documentation when the content has been elided.
- Don't write a source file named after the package itself, this is likely to conflict with resources in the package.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -587,12 +587,14 @@ func cleanupDoc(doc parsedDoc) parsedDoc {
 
 var markdownLink = regexp.MustCompile(`\[([^\]]*)\]\(([^\)]*)\)`)
 
+const elidedDocComment = "<elided>"
+
 // cleanupText processes markdown strings from TF docs and cleans them for inclusion in Pulumi docs
 func cleanupText(text string) string {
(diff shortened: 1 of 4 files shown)


ellismg:
CI is failing because Azure is failing to build, since some of my workarounds for the filename bug are no longer needed.

ellismg:
The source diffs show some comments being dropped for strange reasons I do not fully understand yet.  I still think we want to take these changes because they are an overall improvement.  I filed https://github.com/pulumi/pulumi-terraform/issues/410 to track the open issue.

ellismg:
The build failures in azure are due to the d69ef52.  This is actually the right thing to do.  We just need to remove these lines from the `Makefile`'s when we flow this to other repositories:

https://github.com/pulumi/pulumi-azure/blob/master/Makefile#L23-L27
https://github.com/pulumi/pulumi-openstack/blob/master/Makefile#L25-L27

/cc @stack72

stack72:
Sounds good @ellismg - if you are happy let's merge, I can go ahead and update AWS and OpenStack based on this?

ellismg:
> Sounds good @ellismg - if you are happy let's merge, I can go ahead and update AWS and OpenStack based on this?

SGTM.  Tag me on the reviews, and I'll take a look shortly.

PR #412, opened 2019-07-16
Lift properties so our data sources can be invoked synchronously.

CyrusNajmabadi:


pgavlin:
I was wrong about the codegen tests; they do build the generated code. That's what is currently failing.

> Lift properties so our data sources can be invoked synchronously.

This should be "are" rather than "can be", right?

CyrusNajmabadi:
> This should be "are" rather than "can be", right?

i basically meant: "can be invoked synchronously by the user" as opposed to "the user must asynchronously deal with their promise result."

CyrusNajmabadi:
@pgavlin has verbally signed off.  we're going to try this out downstream to make sure it works.

PR #414, opened 2019-07-17
Expanded error message about a missing field value with a config default

### LINKED ISSUE #400: Better error message about a missing field value with a config default
I had a look at possible ways to fix https://github.com/pulumi/pulumi-azure/issues/278, which lead me to this repository.

The case is the following:

1.  Azure Resource Group has to have a `location` property.
2. It can be set explicitly in the program, then that value is applied.
3. It can be omitted in the program, and then the Azure Provider [specifies](https://github.com/pulumi/pulumi-azure/blob/master/resources.go#L389) `DefaultInfo` to be `Config` from `azure:location` Pulumi configuration value.
4. Because of (2), `azure:location` is an optional parameter, so it is quite easy to do neither (2) nor (3). E.g., run one of our examples without exactly following the README guidance.

In case of (4), the following error is shown to the user on preview: `azure:core/resourceGroup:ResourceGroup resource 'resourceGroup' has a problem: "location": required field is not set`.

The error makes them fix it in the program, which is not desired. We want to guide users towards option (3).

**Desired behavior**

If I understand it correctly, tfbridge could identify this situation and come up with a better error message, e.g. `Required field 'location' is not set for the resource 'resourceGroup'. Either set it explicitly or configure it with 'pulumi config set azure:location <value>`.

Tfbridge seems to have all the information [at this point](https://github.com/pulumi/pulumi-terraform/blob/master/pkg/tfbridge/schema.go#L173), and the behavior appears to be applicable universally.

mikhailshilkov:
Fixes #400 

I'm not very happy about the fact that I had to parse the error message coming from TF and put the logic of checking the config default very far from where it already resides (`MakeTerraformInputs`). However, I don't see another place where we can reliably inject into the pipeline and have no false-positives.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ This CHANGELOG details important changes made in each version of the
 
 - Terraform-based providers can now communicate detailed information about the difference between a resource's desired and actual state during a Pulumi update.
 - Add the ability to inject CustomTimeouts into the InstanceDiff during a pulumi update.
-- Better error message for missing required fields with default config (#400).
+- Better error message for missing required fields with default config ([#400](https://github.com/pulumi/pulumi-terraform/issues/400)).
 
 ## v0.18.3 (Released June 20, 2019)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This CHANGELOG details important changes made in each version of the
 
 - Terraform-based providers can now communicate detailed information about the difference between a resource's desired and actual state during a Pulumi update.
 - Add the ability to inject CustomTimeouts into the InstanceDiff during a pulumi update.
+- Better error message for missing required fields with default config (#400).
 
 ## v0.18.3 (Released June 20, 2019)
 
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -420,13 +420,15 @@ func (p *Provider) Configure(ctx context.Context,
 var requiredFieldRegex = regexp.MustCompile("\"(.*?)\": required field is not set")
 
 func (p *Provider) formatFailureReason(res Resource, reason string) string {
-	// If a required field is missing and the value can be set via config, extend the error with a hint to set the proper config value
+	// If a required field is missing and the value can be set via config,
+	// extend the error with a hint to set the proper config value
 	name := requiredFieldRegex.FindStringSubmatch(reason)
 	if len(name) == 2 {
 		field := res.Schema.Fields[name[1]]
 		if field != nil && field.Default != nil {
 			if configKey := field.Default.Config; configKey != "" {
-				return fmt.Sprintf("%s. Either set it explicitly or configure it with 'pulumi config set %s:%s <value>'.", reason, p.module, configKey)
+				hint := "%s. Either set it explicitly or configure it with 'pulumi config set %s:%s <value>'."
+				return fmt.Sprintf(hint, reason, p.module, configKey)
 			}
 		}
 	}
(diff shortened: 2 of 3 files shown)


mikhailshilkov:
I'm not sure why the downstream generation fails with `Property 'utils' does not exist on type 'typeof import("/home/travis/gopath/src/github.com/pulumi/pulumi-azure/sdk/nodejs/node_modules/@pulumi/pulumi/index")'.`...

Hopefully, not my fault. I did test with Azure, here is a sample improved error message:

```
error: azure:core/resourceGroup:ResourceGroup resource 'resourceGroup' has a problem: "location": 
required field is not set. Either set it explicitly or configure it with 
'pulumi config set azure:location <value>'.
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This CHANGELOG details important changes made in each version of the
 
 - Terraform-based providers can now communicate detailed information about the difference between a resource's desired and actual state during a Pulumi update.
 - Add the ability to inject CustomTimeouts into the InstanceDiff during a pulumi update.
+- Better error message for missing required fields with default config (#400).
```
lukehoban: I *think* this needs to be `[#400](https://github.com/pulumi/pulumi-terraform/issues/400)` to expand correctly?

pkg/tfbridge/provider.go
```diff
@@ -416,6 +417,25 @@ func (p *Provider) Configure(ctx context.Context,
 	return &pulumirpc.ConfigureResponse{}, nil
 }
 
+var requiredFieldRegex = regexp.MustCompile("\"(.*?)\": required field is not set")
```
lukehoban: Perhaps a reference to the string defined here which this regexp should be kept in sync with: https://github.com/hashicorp/terraform/blob/7f5ffbfe9027c34c4ce1062a42b6e8d80b5504e0/helper/schema/schema.go#L1356.

pkg/tfbridge/provider.go
```diff
@@ -416,6 +417,25 @@ func (p *Provider) Configure(ctx context.Context,
...
+
+func (p *Provider) formatFailureReason(res Resource, reason string) string {
+	// If a required field is missing and the value can be set via config,
+	// extend the error with a hint to set the proper config value
+	name := requiredFieldRegex.FindStringSubmatch(reason)
+	if len(name) == 2 {
+		field := res.Schema.Fields[name[1]]
+		if field != nil && field.Default != nil {
+			if configKey := field.Default.Config; configKey != "" {
```
lukehoban: Very nice to see this can be handled specifically for `Config` defaults - so that this narrowly targets the cases where we introduce this issue in the first place!

pkg/tfbridge/provider.go
```diff
@@ -416,6 +417,25 @@ func (p *Provider) Configure(ctx context.Context,
...
+				hint := "%s. Either set it explicitly or configure it with 'pulumi config set %s:%s <value>'."
+				return fmt.Sprintf(hint, reason, p.module, configKey)
```
lukehoban: Is it the case that in general this error message reports back the TF name instead of the Pulumi name?

It's slightly orthogonal to this issue - but this may be an opportunity to fix the naming in the error message to use the Pulumi name (camelCased).  Happy if we want to do that separately - but we've already matched out the name in the regexp here which is the foundation of how we would do that.

lukehoban:
> I'm not sure why the downstream generation fails with Property 'utils' does not exist on type 'typeof import("/home/travis/gopath/src/github.com/pulumi/pulumi-azure/sdk/nodejs/node_modules/@pulumi/pulumi/index")'....

@CyrusNajmabadi just added this yesterday - so possibly a bad interaction with his changes?

pkg/tfbridge/provider.go
mikhailshilkov: Yes, great point! My current code won't work for multi-word names, so that's related enough for sure. I'm changing the code to translate the property name and report it back as Pulumi name.

PR #424, opened 2019-07-19
Change how Tfgen deals with resources called Index

stack72:
We have an issue that Index.ts is a reserves filename in NodeJS
language generation for the contents of the package. So if the resource
is of type Index, we rename it Index_.ts so that it can be included

This allows us to generate the package for GCP Firestore

```
// *** WARNING: this file was generated by the Pulumi Terraform Bridge (tfgen) Tool. ***
// *** Do not edit by hand unless you're certain you know what you are doing! ***

// Export members:
export * from "./index_";
```

This was broken before and we could not use Firestore Indexes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This CHANGELOG details important changes made in each version of the
 
 - Terraform-based providers can now communicate detailed information about the difference between a resource's desired and actual state during a Pulumi update.
 - Add the ability to inject CustomTimeouts into the InstanceDiff during a pulumi update.
+- Change how Tfgen deals with package classes that are named Index to make them index_.ts
 
 ## v0.18.3 (Released June 20, 2019)
 
--- a/pkg/tfgen/generate_nodejs.go
+++ b/pkg/tfgen/generate_nodejs.go
@@ -475,7 +475,16 @@ func (g *nodeJSGenerator) emitPlainOldType(w *tools.GenWriter, pot *plainOldType
 func (g *nodeJSGenerator) emitResourceType(mod *module, res *resourceType) (string, error) {
 	// Create a resource module file into which all of this resource's types will go.
 	name := res.name
-	w, file, err := g.openWriter(mod, lowerFirst(name)+".ts", true)
+	filename := lowerFirst(name)
+
+	// We need to check if the resource is called index or utilities. If so then we will have problems
+	// based on the fact that we need to generate an index.ts based on the package contents
+	// Therefore, we are going to append _ into the name to allow us to continue
+	if filename == "index" || filename == "utilities" {
+		filename = fmt.Sprintf("%s_", filename)
+	}
+
+	w, file, err := g.openWriter(mod, filename+".ts", true)
 	if err != nil {
 		return "", err
 	}


pkg/tfgen/generate_nodejs.go
```diff
@@ -475,7 +475,17 @@ func (g *nodeJSGenerator) emitPlainOldType(w *tools.GenWriter, pot *plainOldType
...
 	// Create a resource module file into which all of this resource's types will go.
 	name := res.name
-	w, file, err := g.openWriter(mod, lowerFirst(name)+".ts", true)
+
+	filename := name
+
+	// We need to check if the resource is called Index. If so then we will have problems
+	// based on the fact that we need to generate an Index.ts based on the package contents
+	// Therefore, we are going to append _ into the name to allow us to continue
+	if filename == "Index" {
```
pgavlin: This check should be case-insensitive.

pkg/tfgen/generate_nodejs.go
```diff
@@ -475,7 +475,17 @@ func (g *nodeJSGenerator) emitPlainOldType(w *tools.GenWriter, pot *plainOldType
...
+		filename = fmt.Sprintf("%s_", name)
+	}
```
lukehoban: It’s logically the same, but seems cleaner to do the lowerFirst first, then compare with the “reserved filenames (`index.ts`, I think there are others?)?

pgavlin: Yes, we should also look for `utilities.ts`

pkg/tfgen/generate_nodejs.go
stack72: Fixed and pushed!

pkg/tfgen/generate_nodejs.go
stack72: Fixed and pushed - also checking for utilities as that is also a reserved filename

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This CHANGELOG details important changes made in each version of the
 
 - Terraform-based providers can now communicate detailed information about the difference between a resource's desired and actual state during a Pulumi update.
 - Add the ability to inject CustomTimeouts into the InstanceDiff during a pulumi update.
+- Change how Tfgen deals with package classes that are named Index to make them Index_.ts
```
justinvp: Nit

```suggestion
- Change how Tfgen deals with package classes that are named Index to make them index_.ts
```

pkg/tfgen/generate_nodejs.go
```diff
@@ -475,7 +475,16 @@ func (g *nodeJSGenerator) emitPlainOldType(w *tools.GenWriter, pot *plainOldType
...
 	// Create a resource module file into which all of this resource's types will go.
 	name := res.name
-	w, file, err := g.openWriter(mod, lowerFirst(name)+".ts", true)
+	filename := lowerFirst(name)
+
+	// We need to check if the resource is called index or utilities. If so then we will have problems
+	// based on the fact that we need to generate an index.ts based on the package contents
+	// Therefore, we are going to append _ into the name to allow us to continue
+	if filename == "index" || filename == "utilities" {
+		filename = fmt.Sprintf("%s_", name)
```
lukehoban: Name should be filename here.

stack72: @lukehoban done!

PR #426, opened 2019-07-23
Use Python Codegen package from pulumi/pulumi

stack72:


pkg/tfgen/generate_python.go
```diff
@@ -23,12 +23,12 @@ import (
 	"sort"
 	"strconv"
 	"strings"
-	"unicode"
 
 	"github.com/golang/glog"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/pkg/errors"
 
+	pyCodeGen "github.com/pulumi/pulumi/pkg/codegen/python"
```
pgavlin: nit: `pycodegen` or `pycode`

pkg/tfgen/generate_python_test.go
```diff
@@ -18,6 +18,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
+	pyCodeGen "github.com/pulumi/pulumi/pkg/codegen/python"
```
pgavlin: Same here

Issue #406, opened 2019-07-11
Treat terraform "sensitive" outputs as Pulumi secrets


ellismg:
During a resource operation a provider is able to communicate information about outputs that should be persisted securely. These values are then treated as secrets by the rest of the system, which means they are encrypted before being stored in the Pulumi state file and the resulting `Output<T>`'s inside the language are also marked as secret.

Today, TF Bridges providers do not communicate back any information about what should be treated as a secret, but we can infer some of these values from the terraform schema, since they will be marked as sensitive.

We should update our logic such that we mark these values as secrets, when the engine supports them.

lukehoban:
This issue presupposes that the Terrraform schema includes specification of "sensitive" outputs.  But as far as I can tell, there is no such thing.  There is https://www.terraform.io/docs/configuration/outputs.html#sensitive-suppressing-values-in-cli-output, but that is only for `output` blocks in terraform, not resource state, so is not what's needed here.

Is this issue not actionable?

If not - what should be done to make something like `Random` easier to use "safely"?

pgavlin:
@lukehoban see https://github.com/hashicorp/terraform/blob/master/helper/schema/schema.go#L245-L249 for the definition of `Sensitive` in the TF schema package.

lukehoban:
Indeed you are right!  Part of my confusion was that `Random` apparently doesn't use this - which is surprising as this is probably one of the primary source for generating sensitive values.   But great to see `Sensitive` is well-used in the AWS provider at least, so there is significant value here.

pgavlin:
We could add our own overlays ala `AltTypes` for cases like Random if we feel it’s valuable enough.

PR #427, opened 2019-07-23
Treat sensitive outputs as secrets

stack72:
Fixes: #406

When an output is marked as sensitive according to the schema, we need
to return this as a Pulumi secret

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -43,15 +43,16 @@ import (
 
 // Provider implements the Pulumi resource provider operations for any Terraform plugin.
 type Provider struct {
-	host         *provider.HostClient               // the RPC link back to the Pulumi engine.
-	module       string                             // the Terraform module name.
-	version      string                             // the plugin version number.
-	tf           *schema.Provider                   // the Terraform resource provider to use.
-	info         ProviderInfo                       // overlaid info about this provider.
-	config       map[string]*schema.Schema          // the Terraform config schema.
-	configValues resource.PropertyMap               // this package's config values.
-	resources    map[tokens.Type]Resource           // a map of Pulumi type tokens to resource info.
-	dataSources  map[tokens.ModuleMember]DataSource // a map of Pulumi module tokens to data sources.
+	host            *provider.HostClient               // the RPC link back to the Pulumi engine.
+	module          string                             // the Terraform module name.
+	version         string                             // the plugin version number.
+	tf              *schema.Provider                   // the Terraform resource provider to use.
+	info            ProviderInfo                       // overlaid info about this provider.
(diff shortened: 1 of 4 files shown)


pkg/tfbridge/schema_test.go
```diff
@@ -325,6 +326,11 @@ func TestTerraformOutputs(t *testing.T) {
 					},
 				},
 			},
+			"secret_value": {
```
pgavlin: Can you add a test for a nested secret property?

stack72: Sure! Will do :)

stack72: done!

stack72:
@pgavlin I added a test to show that when the secrets are not supported from the provider that we don't treat as sensitive

pkg/tfbridge/provider.go
```diff
@@ -43,15 +43,16 @@ import (
...
+	host            *provider.HostClient               // the RPC link back to the Pulumi engine.
+	module          string                             // the Terraform module name.
+	version         string                             // the plugin version number.
+	tf              *schema.Provider                   // the Terraform resource provider to use.
+	info            ProviderInfo                       // overlaid info about this provider.
+	config          map[string]*schema.Schema          // the Terraform config schema.
+	configValues    resource.PropertyMap               // this package's config values.
+	resources       map[tokens.Type]Resource           // a map of Pulumi type tokens to resource info.
+	dataSources     map[tokens.ModuleMember]DataSource // a map of Pulumi module tokens to data sources.
+	supportsSecrets bool                               //understanding whether the provider supports secrets
```
pgavlin:
```suggestion
	supportsSecrets bool                               // true if the engine supports secret property values
```

pkg/tfbridge/schema.go
```diff
@@ -524,122 +526,133 @@ func MakeTerraformOutputs(outs map[string]interface{},
...
-		}
-		var tfes *schema.Schema
-		if tfs != nil {
-			if sch, issch := tfs.Elem.(*schema.Schema); issch {
-				tfes = sch
-			} else if _, isres := tfs.Elem.(*schema.Resource); isres {
-				// The map[string]interface{} case below expects a schema whose
-				// `Elem` is a Resource, so just pass the full List schema
-				tfes = tfs
+	tfs *schema.Schema, ps *SchemaInfo, assets AssetTable, rawNames bool, supportsSecrets bool) resource.PropertyValue {
```
pgavlin:
```suggestion
	tfs *schema.Schema, ps *SchemaInfo, assets AssetTable, rawNames, supportsSecrets bool) resource.PropertyValue {
```

pgavlin: (likewise in `MakeTerraformOutputs`)

PR #434, opened 2019-07-26
Add support in pulumi-terraform for providing aliases for a particular resource.

CyrusNajmabadi:
Will be used downstream initially in pulumi/aws and pulumi/azure to rename some resources in a non-destructive way for users.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -84,7 +84,7 @@ type ResourceInfo struct {
 	IDFields            []string               // an optional list of ID alias fields.
 	Docs                *DocInfo               // overrides for finding and mapping TF docs.
 	DeleteBeforeReplace bool                   // if true, Pulumi will delete before creating new replacement resources.
-	Aliases             []AliasInfo            // sliases for this resources, if any.
+	Aliases             []AliasInfo            // aliases for this resources, if any.
 }
 
 // DataSourceInfo can be used to override a data source's standard name mangling and argument/return information.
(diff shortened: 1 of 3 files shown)


pkg/tfbridge/info.go
```diff
@@ -72,6 +84,7 @@ type ResourceInfo struct {
 	IDFields            []string               // an optional list of ID alias fields.
 	Docs                *DocInfo               // overrides for finding and mapping TF docs.
 	DeleteBeforeReplace bool                   // if true, Pulumi will delete before creating new replacement resources.
+	Aliases             []AliasInfo            // sliases for this resources, if any.
```
CyrusNajmabadi:
```suggestion
	Aliases             []AliasInfo            // aliases for this resources, if any.
```

stack72:
Hi @CyrusNajmabadi 	

I love the idea of this - it's going to work well with the work I am planning on adding the ability to add a deprecation message in pulumi-terraform for resources

Can you show an example of how this is used?

P.

CyrusNajmabadi:
> it's going to work well with the work I am planning on adding the ability to add a deprecation message in pulumi-terraform for resources

That's fantastic.  I need that as well :)   So it sounds like our work compliments each other well.

See https://github.com/pulumi/pulumi-aws/pull/675/files#diff-3d6b5e2ce37c7b5b2214e0a7d75ec1d7R34 to see how this would be used in practice.

CyrusNajmabadi:
Have validated this works downstream in pulumi-aws

CyrusNajmabadi:
I'm going to move forward on this.  If tehre are concerns that you'd like addressed, let me know!

stack72:
FWIW, I am happy with this as well - I had to change locations before posting! nice work here @CyrusNajmabadi

Issue #392, opened 2019-06-18
Generated setup.py's assume `pulumi` is installed and on the $PATH


ellismg:
As part of generating a setup.py for a package, we include code that runs at install time and calls `pulumi plugin install` in order to install the corresponding resource plugin.

https://github.com/pulumi/pulumi-terraform/blob/7ace3e9b5f2dcd4692b029ba4b80360582d7949a/pkg/tfgen/generate_python.go#L642

When `pulumi` is not on the $PATH, this call will fail, which will cause `check_call` to throw an exception which will be printed during package install.  We should do something similar to what we do in node where we check to see if `pulumi` is on the `$PATH` and if it is not, we instead warn the user that they will need to manually run a `pulumi plugin install` command after installing the tool.

PR #435, opened 2019-07-28
Emit warning that pulumi binary not found when installing plugin

stack72:
Fixes: #392

We don't want to raise an exception to the user, instead, we want
to warn the user that pulumi binary isn't found and want them to
install the plugin by hand after pulumi is installed

### DIFF
--- a/pkg/tfgen/generate_python.go
+++ b/pkg/tfgen/generate_python.go
@@ -721,7 +721,7 @@ func (g *pythonGenerator) emitPackageMetadata(pack *pkg) error {
 	w.Writefmtln("                There was an error installing the %s resource provider plugin.", pack.name)
 	w.Writefmtln("                It looks like `pulumi` is not installed on your system.")
 	w.Writefmtln("                Please visit https://pulumi.com/ to get Pulumi.")
-	w.Writefmtln("                You may try to manually installing the plugin by running")
+	w.Writefmtln("                You may try manually installing the plugin by running")
 	w.Writefmtln("                `pulumi plugin install resource %s ${PLUGIN_VERSION}`", pack.name)
 	w.Writefmtln("                \"\"\")")
 	w.Writefmtln("            else:")
--- a/pkg/tfgen/generate_python.go
+++ b/pkg/tfgen/generate_python.go
@@ -720,7 +720,7 @@ func (g *pythonGenerator) emitPackageMetadata(pack *pkg) error {
 	w.Writefmtln("                print(\"\"\"")
 	w.Writefmtln("                There was an error installing the %s resource provider plugin.", pack.name)
 	w.Writefmtln("                It looks like `pulumi` is not installed on your system.")
-	w.Writefmtln("                Please visit https://pulumi.com/ to get Pulumi.")
+	w.Writefmtln("                Please visit https://pulumi.com/ to install the Pulumi CLI.")
 	w.Writefmtln("                You may try manually installing the plugin by running")
 	w.Writefmtln("                `pulumi plugin install resource %s ${PLUGIN_VERSION}`", pack.name)
 	w.Writefmtln("                \"\"\")")
--- a/pkg/tfgen/generate_python.go
+++ b/pkg/tfgen/generate_python.go
@@ -702,6 +702,7 @@ func (g *pythonGenerator) emitPackageMetadata(pack *pkg) error {
 	w.EmitHeaderWarning(g.commentChars())
 
 	// Now create a standard Python package from the metadata.
+	w.Writefmtln("import errno")
 	w.Writefmtln("from setuptools import setup, find_packages")
 	w.Writefmtln("from setuptools.command.install import install")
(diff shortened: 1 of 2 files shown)


stack72:
FYI, OSError is the better way to handle this than scanning in the search path - that can be a costly event!

stack72:
@ellismg ok, after testing this, I not get the following error when Pulumi isn't found:

Screenshot 2019-07-31 at 09 37 45

I think this follows what we do in node-land

stack72:
I expect the Azure downstream build to fail until - https://github.com/pulumi/pulumi-azure/pull/320 is merged

pkg/tfgen/generate_python.go
```diff
@@ -711,8 +712,20 @@ func (g *pythonGenerator) emitPackageMetadata(pack *pkg) error {
...
+	w.Writefmtln("            check_call(['pulumi', 'plugin', 'install', 'resource', '%s', '${PLUGIN_VERSION}'])",
 		pack.name)
+	w.Writefmtln("        except OSError as error:")
+	w.Writefmtln("            if error.errno == errno.ENOENT:")
+	w.Writefmtln("                print(\"\"\"")
+	w.Writefmtln("                There was an issue installing the resource provider plugin.")
+	w.Writefmtln("                It looks like `pulumi` is not installed on your system.")
+	w.Writefmtln("                Please visit https://pulumi.io/ to get Pulumi.")
+	w.Writefmtln("                You may try to manually installing the plugin by running")
+	w.Writefmtln("                `plugin install resource %s ${PLUGIN_VERSION}`", pack.name)
```
jen20: This should be `pulumi plugin install`, no?

pkg/tfgen/generate_python.go
```diff
@@ -711,8 +712,20 @@ func (g *pythonGenerator) emitPackageMetadata(pack *pkg) error {
...
 	w.Writefmtln("    def run(self):")
 	w.Writefmtln("        install.run(self)")
-	w.Writefmtln("        check_call(['pulumi', 'plugin', 'install', 'resource', '%s', '${PLUGIN_VERSION}'])",
+	w.Writefmtln("        try:")
+	w.Writefmtln("            check_call(['pulumi', 'plugin', 'install', 'resource', '%s', '${PLUGIN_VERSION}'])",
 		pack.name)
+	w.Writefmtln("        except OSError as error:")
+	w.Writefmtln("            if error.errno == errno.ENOENT:")
+	w.Writefmtln("                print(\"\"\"")
+	w.Writefmtln("                There was an issue installing the resource provider plugin.")
```
jen20: s/issue/error/

jen20: We should probably mention the provider name here too, so: "There was an error installing the X resource provider plugin."

stack72: This is copied exactly from the NodeJS version...

pkg/tfgen/generate_python.go
stack72: good catch!

pkg/tfgen/generate_python.go
stack72: Updated and also changed to pulumi.com rather than pulumi.io

pkg/tfgen/generate_python.go
```diff
@@ -711,8 +712,20 @@ func (g *pythonGenerator) emitPackageMetadata(pack *pkg) error {
...
-	w.Writefmtln("        check_call(['pulumi', 'plugin', 'install', 'resource', '%s', '${PLUGIN_VERSION}'])",
+	w.Writefmtln("        try:")
+	w.Writefmtln("            check_call(['pulumi', 'plugin', 'install', 'resource', '%s', '${PLUGIN_VERSION}'])",
 		pack.name)
+	w.Writefmtln("        except OSError as error:")
+	w.Writefmtln("            if error.errno == errno.ENOENT:")
+	w.Writefmtln("                print(\"\"\"")
+	w.Writefmtln("                There was an error installing the %s resource provider plugin.", pack.name)
+	w.Writefmtln("                It looks like `pulumi` is not installed on your system.")
+	w.Writefmtln("                Please visit https://pulumi.com/ to get Pulumi.")
```
pgavlin:
```suggestion
	w.Writefmtln("                Please visit https://pulumi.com/ to install the Pulumi CLI.")
```

pkg/tfgen/generate_python.go
```diff
@@ -711,8 +712,20 @@ func (g *pythonGenerator) emitPackageMetadata(pack *pkg) error {
...
+	w.Writefmtln("                You may try to manually installing the plugin by running")
```
pgavlin:
```suggestion
	w.Writefmtln("                You may try manually installing the plugin by running")
```

pgavlin:
I see that the text you have matches the text @ellismg linked to, so feel free to accept or ignore my suggestions at your leisure.

PR #437, opened 2019-07-29
Add the ability to add a deprecation notice to resources

stack72:
Based on being able to add a DeprecationMessage to the schema, e.g. :

```
"aws_s3_bucket": {
	DeprecationMessage: "use aws_s3_new_bucket instead",
	Tok:                awsResource(s3Mod, "Bucket"),
},
```

We will emit Deprecation decorators across JavaScript, Go and Python langague SDKs. 

As a bonus, for JavaScript and Python, we can get outputs as follows:

```TypeScript
..[$] <()> pulumi up
Previewing update (dev):

     Type                 Name                            Plan       Info
 +   pulumi:pulumi:Stack  aws-ts-deprecation-testing-dev  create     1 warning
 +   └─ aws:s3:Bucket     my-bucket                       create

Diagnostics:
  pulumi:pulumi:Stack (aws-ts-deprecation-testing-dev):
    warning: Bucket is deprecated: use aws_s3_new_bucket instead

Resources:
    + 2 to create
```

```Python
$ pulumi up
Previewing update (dev):

     Type                 Name                            Plan       Info
 +   pulumi:pulumi:Stack  aws-py-deprecation-testing-dev  create     1 warning
 +   └─ aws:s3:Bucket     my-bucket                       create

Diagnostics:
  pulumi:pulumi:Stack (aws-py-deprecation-testing-dev):
    warning: bucket is deprecated: use aws_s3_new_bucket instead

Resources:
    + 2 to create
```

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -85,13 +85,15 @@ type ResourceInfo struct {
 	Docs                *DocInfo               // overrides for finding and mapping TF docs.
 	DeleteBeforeReplace bool                   // if true, Pulumi will delete before creating new replacement resources.
 	Aliases             []AliasInfo            // aliases for this resources, if any.
+	DeprecationMessage  string                 // message to use in deprecation warning
 }
 
 // DataSourceInfo can be used to override a data source's standard name mangling and argument/return information.
 type DataSourceInfo struct {
-	Tok    tokens.ModuleMember
-	Fields map[string]*SchemaInfo
-	Docs   *DocInfo // overrides for finding and mapping TF docs.
+	Tok                tokens.ModuleMember
+	Fields             map[string]*SchemaInfo
+	Docs               *DocInfo // overrides for finding and mapping TF docs.
+	DeprecationMessage string   // message to use in deprecation warning
 }
 
 // SchemaInfo contains optional name transformations to apply.
--- a/pkg/tfgen/generate_go.go
+++ b/pkg/tfgen/generate_go.go
@@ -438,6 +438,11 @@ func (g *goGenerator) emitResourceType(mod *module, res *resourceType) error {
 	if res.doc != "" {
 		g.emitDocComment(w, res.doc, res.docURL, "")
 	}
+	if !res.IsProvider() {
+		if res.info.DeprecationMessage != "" {
+			w.Writefmtln("// Deprecated: %s", res.info.DeprecationMessage)
+		}
+	}
 	w.Writefmtln("type %s struct {", name)
 	w.Writefmtln("\ts *pulumi.ResourceState")
 	w.Writefmtln("}")
@@ -446,6 +451,9 @@ func (g *goGenerator) emitResourceType(mod *module, res *resourceType) error {
 	// Create a constructor function that registers a new instance of this resource.
 	argsType := res.argst.name
 	w.Writefmtln("// New%s registers a new resource with the given unique name, arguments, and options.", name)
+	if res.info.DeprecationMessage != "" {
+		w.Writefmtln("// Deprecated: %s", res.info.DeprecationMessage)
+	}
 	w.Writefmtln("func New%s(ctx *pulumi.Context,", name)
 	w.Writefmtln("\tname string, args *%s, opts ...pulumi.ResourceOpt) (*%s, error) {", argsType, name)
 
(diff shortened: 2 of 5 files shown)


pkg/tfbridge/info.go
```diff
@@ -85,6 +85,8 @@ type ResourceInfo struct {
 	Docs                *DocInfo               // overrides for finding and mapping TF docs.
 	DeleteBeforeReplace bool                   // if true, Pulumi will delete before creating new replacement resources.
 	Aliases             []AliasInfo            // aliases for this resources, if any.
+	Deprecated          bool                   // if true, the resource will be marked as deprecated
+	Message             string                 // message to use in deprecation warning
```
CyrusNajmabadi: Consider calling this DeprecationMessage.

pkg/tfgen/generate_go.go
```diff
@@ -438,6 +438,12 @@ func (g *goGenerator) emitResourceType(mod *module, res *resourceType) error {
 	if res.doc != "" {
 		g.emitDocComment(w, res.doc, res.docURL, "")
 	}
+	if !res.IsProvider(){
+		deprecationMessage := g.getDeprecationDetails(res.info)
+		if deprecationMessage != "" {
+			w.Writefmtln("%s", deprecationMessage)
```
CyrusNajmabadi: my thinking on deprecation is that it should override normal docs.  so we should just check this first and skip normal docs if we have it.

pkg/tfgen/generate_python.go
```diff
@@ -457,6 +457,17 @@ func (g *pythonGenerator) emitResourceType(mod *module, res *resourceType) (stri
...
+		deprecationMessage := g.getDeprecationDetails(res.info)
+		if deprecationMessage != "" {
+			w.Writefmtln("warnings.warn(")
+			w.Writefmtln("    %s,", deprecationMessage)
+			w.Writefmtln("    DeprecationWarning")
+			w.Writefmtln(")")
+		}
+	}
+
 	// Produce a class definition with optional """ comment.
```
CyrusNajmabadi: avoid comment if the type is deprecated.  or, the comment should just have the deprecation message in it.

pkg/tfbridge/info.go
lukehoban: Also - can we just combine these two fields down to just one?

stack72: Happy to make the change - just pushed it - if we have a `DeprecationMessage` we are now marking something as `Deprecated` no need for the 2

pkg/tfgen/generate_nodejs.go
```diff
@@ -1078,6 +1090,14 @@ func (g *nodeJSGenerator) gatherCustomImports(mod *module, info *tfbridge.Schema
 	return nil
 }
 
+func (g *nodeJSGenerator) getDeprecationDetails(message string) string {
+	if message != "" {
+		return fmt.Sprintf("/** @deprecated %s%s*/", message, " ")
```
lukehoban: Why the second `%s` for `" "`?

lukehoban: Any reason not to just inline this?  We're effectively just doing the `if` twice in all the locations above - right?

pkg/tfgen/generate_python.go
```diff
@@ -616,6 +626,13 @@ func (g *pythonGenerator) emitResourceFunc(mod *module, fun *resourceFunc) (stri
 	}
 	defer contract.IgnoreClose(w)
 
+	if fun.info.DeprecationMessage != "" {
+		w.Writefmtln("warnings.warn(")
+		w.Writefmtln("    \"%s\",", fun.info.DeprecationMessage)
+		w.Writefmtln("    DeprecationWarning")
+		w.Writefmtln(")")
```
lukehoban: Any particular reason we are splitting this over 4 lines of generated code?  Seems it may as well just be one.

lukehoban:
This doesn't look right - won't that issue the warning when the package is loaded, instead when the class is instantiated?

```py
warnings.warn(
    "This resource is deprecated - this is a test message only",
    DeprecationWarning
)
class Queue(pulumi.CustomResource):
```

lukehoban:
Curious - what exactly is the effect this has?  Does it warn on trying to call the constructor or do we need to also emit the deprecation there as well?

```ts
/** @deprecated This resource is deprecated - this is a test message only */
export class Queue extends pulumi.CustomResource {
```

pkg/tfgen/generate_nodejs.go
stack72: done!

pkg/tfgen/generate_python.go
stack72: This is just to ensure that PEP3 doesn't get tripped up when we have a longer message

pkg/tfgen/generate_go.go
```diff
@@ -438,6 +438,11 @@ func (g *goGenerator) emitResourceType(mod *module, res *resourceType) error {
 	if res.doc != "" {
 		g.emitDocComment(w, res.doc, res.docURL, "")
 	}
+	if !res.IsProvider() {
```
pgavlin:
```suggestion
	if !res.IsProvider() && res.info.DeprecationMessage != "" {
```

pgavlin: (same throughout)

stack72: Fixed up!

pgavlin: Looks like this didn't actually get changed?

stack72: all changed!

```
	if !res.IsProvider() && res.info.DeprecationMessage != "" {
		w.Writefmtln("warnings.warn(\"%s\", DeprecationWarning)", res.info.DeprecationMessage)
	}
~``

PR #431, opened 2019-07-24
Enabling the ability for the Marshalling of Secrets in outputs

stack72:
Part of #427 	

Via CLI:
```
Previewing update (dev):

     Type                  Name                   Plan
 +   pulumi:pulumi:Stack   encrypted-secrets-dev  create
 +   └─ aws:ssm:Parameter  foo                    create

Resources:
    + 2 to create

Updating (dev):

     Type                  Name                   Status
 +   pulumi:pulumi:Stack   encrypted-secrets-dev  created
 +   └─ aws:ssm:Parameter  foo                    created

Outputs:
    value: "[secret]"

Resources:
    + 2 created

Duration: 7s
```

Via checkpoint file
```
{
                "urn": "urn:pulumi:dev::encrypted-secrets::aws:ssm/parameter:Parameter::foo",
                "custom": true,
                "id": "foo-58ce001",
                "type": "aws:ssm/parameter:Parameter",
                "inputs": {
                    "__defaults": [
                        "name",
                        "tier"
                    ],
                    "name": "foo-58ce001",
                    "tier": "Standard",
                    "type": "SecureString",
                    "value": {
                        "4dabf18193072939515e22adb298388d": "1b47061264138c4ac30d75fd1eb44270",
                        "ciphertext": "AAABAHHL2iJu4XKEaaguQnS/RvJcP+XlfY5MwstlxxdYTvC72Q=="
                    }
                },
                "outputs": {
                    "allowedPattern": "",
                    "arn": "arn:aws:ssm:us-east-1:153052954103:parameter/foo-58ce001",
                    "description": "",
                    "id": "foo-58ce001",
                    "keyId": "alias/aws/ssm",
                    "name": "foo-58ce001",
                    "tags": {},
                    "tier": "Standard",
                    "type": "SecureString",
                    "value": {
                        "4dabf18193072939515e22adb298388d": "1b47061264138c4ac30d75fd1eb44270",
                        "ciphertext": "AAABALrta9Syc716HgRzeDb0TH1KMlimgztuPiGUNOWPr4q0Ig=="
                    },
                    "version": 1
                },
                "parent": "urn:pulumi:dev::encrypted-secrets::pulumi:pulumi:Stack::encrypted-secrets-dev",
                "provider": "urn:pulumi:dev::encrypted-secrets::pulumi:providers:aws::default::c2915197-f245-4d96-b068-97479a871a8d",
                "propertyDependencies": {
                    "type": null,
                    "value": null
                },
                "customTimeouts": {
[...]

PR #439, opened 2019-07-30
Revert "Enabling the ability for the Marshalling of Secrets in outputs"

stack72:
Reverts pulumi/pulumi-terraform#431


stack72:
Breaks my heart but it's understandable

PR #440, opened 2019-07-30
Upgrade deps to Terraform v0.12.5

stack72:
This was required because config.UnknownVariableValue has
moved packages to become hcl2shim.UnknownVariableValue


stack72:
The Azure build is expected to fail here - that provider references `config.UnknownVariableValue` directly in resources.go and will need to be upgraded itself

PR #441, opened 2019-07-30
Support `ignoreChanges.`

pgavlin:
These changes implement the provider side of the `ignoreChanges` support
added in https://github.com/pulumi/pulumi/pull/3005. This support is
implemented by crawling the old state and new config to map the set
of ignored paths to TF attribute keys, then removing any changes to
these keys from the diff in question.


pgavlin:
NOTE: this will fail until https://github.com/pulumi/pulumi/pull/3005 is merged and this repo's `go.mod` is updated.

pkg/tfbridge/diff.go
```diff
@@ -47,7 +47,9 @@ func containsComputedValues(v resource.PropertyValue) bool {
 	}
 }
 
-// makePropertyDiff checks the given property for diffs and records any that are found in diff.
+type propertyVisitor func(attributeKey, propertyPath string, value resource.PropertyValue) bool
```
ellismg: Could you comment what the semantics of the return value are of this function?

pgavlin: Done. Thank you :)

PR #442, opened 2019-07-31
Re-enable the provider version pinning code for Python

stack72:


stack72:
No changelog as it's covered by another PR already merged!

PR #443, opened 2019-07-31
Update pulumi/pulumi dependency

jen20:
This brings the `pulumi/pulumi` dependency up to `master`, allowing us to remove the `replace` blocks in the `go.mod` file.


pgavlin:
This is no longer necessary due to the changes in https://github.com/pulumi/pulumi-terraform/pull/441/files#diff-37aff102a57d3d7b797f152915a6dc16.

PR #444, opened 2019-08-01
Take schema.StateFunc into account when propagating defaults.

pgavlin:
String properties may be normalized by schema.StateFunc before they are
stored into TF state. When this happens for properties that are commonly
given default values (e.g. `name` properties), this normalization can
cause `Read`'s default propagation to consider the property changed and
drop the default value. This in turn can lead to spurious diffs when a
new default is calculated during the next update.

In order to fix this, normalize any inputs with a schema.StateFunc
before comparing them for equality. Note that only `string`-typed
 properties support `schema.StateFunc`.

Fixes https://github.com/pulumi/pulumi-aws/issues/684.


stack72:
@pgavlin I think this also fixes https://github.com/pulumi/pulumi-terraform/issues/408 right?

pgavlin:
So it does!

Issue #384, opened 2019-06-04
Add a Python package to expose RemoteStateReference


jen20:
Currently the work to expose Terraform remote states only works from Node.js, as it was blocked on Read support in Python. Now that exists, we can publish a Python equivalent of the `@pulumi/pulumi` package exposing this.

lukehoban:
Addressed with https://github.com/pulumi/pulumi-terraform/pull/445.

PR #445, opened 2019-08-02
Add a Python SDK for pulumi-terraform

jen20:
This pull request adds an SDK for `pulumi-terraform` to allow consuming remote state resources from Python. Note that since `pulumi-terraform` does not use `tfbridge`, these types are hand-written (similar to the Node.js SDK) rather than generated.

TODO:

- [x] SDK structure
- [x] Types for RemoteStateReference and arguments
- [x] Test using `backend_type = "local"` as per the Node.js SDK
- [x] Update build system to build and publish packages.


jen20:
Fixes #384.

jen20:
The failure here is downstream but unrelated to this PR.

PR #447, opened 2019-08-05
Take advantage of synchronous invokes in Python.

pgavlin:
With these changes, all invokes will now run synchronously. We retain
backwards compatibility with user programs by making the result of each
invoke awaitable.

Because synchronous invokes require that bridged providers depend on a
newer version of the Pulumi Python SDK, these changes also introduce
some basic validation of the Python requirements entry in the provider
info.

### DIFF
--- a/pkg/tfgen/generate_python.go
+++ b/pkg/tfgen/generate_python.go
@@ -20,10 +20,12 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
 
+	"github.com/blang/semver"
 	"github.com/golang/glog"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


pgavlin:
Codegen jobs are expected to fail as all downstream repos need a Python SDK version bump.

pkg/tfgen/generate_python.go
```diff
@@ -634,10 +636,23 @@ func (g *pythonGenerator) emitResourceFunc(mod *module, fun *resourceFunc) (stri
 	if fun.retst != nil {
 		g.emitPlainOldType(w, fun.retst)
 		w.Writefmtln("")
+
+		// Emit __await__ and __iter__ in order to make this type awaitable.
+		//
+		// Note that we need __await__ to be an iterator, but we only want it to return one value. As such, we use
+		// `if False: yield` to construct this.
```
CyrusNajmabadi: thanks for the comment. that was a headscratcher

PR #448, opened 2019-08-06
Rewrite resource and property names in doc generator

lukehoban:
Part of #398.

See:
- GCP: https://github.com/pulumi/pulumi-gcp/compare/lukehoban/namerewriting
- AWS: https://github.com/pulumi/pulumi-aws/compare/lukehoban/namerewriting
- Azure: https://github.com/pulumi/pulumi-azure/compare/lukehoban/namerewriting

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -144,8 +144,9 @@ func readMarkdown(repo string, kind DocKind, possibleLocations []string) ([]byte
 }
 
 // mergeDocs adds the docs specified by extractDoc from sourceFrom into the targetDocs
-func mergeDocs(g *generator, info tfbridge.ResourceOrDataSourceInfo, org string, provider string, resourcePrefix string, kind DocKind,
-	targetDocs map[string]string, sourceFrom string, extractDocs func(d parsedDoc) map[string]string) error {
+func mergeDocs(g *generator, info tfbridge.ResourceOrDataSourceInfo, org string, provider string,
+	resourcePrefix string, kind DocKind, targetDocs map[string]string, sourceFrom string,
+	extractDocs func(d parsedDoc) map[string]string) error {
 
 	if sourceFrom != "" {
 		sourceDocs, err := getDocsForProvider(g, org, provider, resourcePrefix, kind, sourceFrom, nil)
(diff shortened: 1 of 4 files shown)


pkg/tfgen/docs.go
```diff
@@ -621,6 +622,35 @@ func cleanupText(text string) string {
...
+		if hasResourceInfo {
+			// This is a resource name
+			resname, mod := resourceName(g.info.GetResourcePrefix(), name, info, false)
+			modname := extractModuleName(mod)
+			switch g.language {
+			case golang, python:
+				// Use `ec2.Instance` format
+				return parts[1] + modname + "." + resname + parts[4]
+			default:
+				// User `aws.ec2.Instance` format`
```
justinvp: Nit:

```suggestion
				// Use `aws.ec2.Instance` format`
```

pkg/tfgen/docs.go
```diff
@@ -621,6 +622,35 @@ func cleanupText(text string) string {
...
+				return parts[1] + g.info.GetResourcePrefix() + "." + modname + "." + resname + parts[4]
```
justinvp: Using `g.info.GetResourcePrefix()` returns `google` for the GCP provider, but we generally refer to these using the package name, e.g. `gcp`.

For example:

https://github.com/pulumi/pulumi-gcp/compare/lukehoban/namerewriting#diff-5ecf084a4a91a5264583a5959a8a5f04R125

`google.compute.Image` should be `gcp.compute.Image`.

lukehoban: Fixed.

pkg/tfgen/docs.go
lukehoban: Fixed.

lukehoban:
> The following is referring to a data source as googleIamPolicy which should be IAMPolicy

Indeed - this wasn't handling datasources previously - but now is.

> this is changing string values in examples, but all the cases I saw seem benign (and in many cases, desirable, to match the associated variable name). Ideally, these would be fixed-up in tf2pulumi rather than this docs cleanup pass, but it's a nice side effect for the time being.

Yeah - it seems like this does result in a net improvement even though it wasn't explicitly part of the goal.

FWIW - I started out looking just at words surrounded by backticks, but there were so many references to property names outside of backticks that this felt very uneven.

PR #450, opened 2019-08-06
Cleanup the code we generate for python resources.

CyrusNajmabadi:


pkg/tfgen/generate_python.go
```diff
@@ -493,12 +493,6 @@ func (g *pythonGenerator) emitResourceType(mod *module, res *resourceType) (stri
 	w.Writefmtln(
 		`            warnings.warn("explicit use of __opts__ is deprecated, use 'opts' instead", DeprecationWarning)`)
 	w.Writefmtln("            opts = __opts__")
-	w.Writefmtln("        if not resource_name:")
-	w.Writefmtln("            raise TypeError('Missing resource name argument (for URN creation)')")
-	w.Writefmtln("        if not isinstance(resource_name, str):")
-	w.Writefmtln("            raise TypeError('Expected resource name to be a string')")
-	w.Writefmtln("        if opts and not isinstance(opts, pulumi.ResourceOptions):")
-	w.Writefmtln("            raise TypeError('Expected resource options to be a ResourceOptions instance')")
```
CyrusNajmabadi: not needed.  We already do this in `class Resource:`:

```
        if not name:
            raise TypeError(
                'Missing resource name argument (for URN creation)')
        if not isinstance(name, str):
            raise TypeError('Expected resource name to be a string')
```

Going to add the ` isinstance(opts, pulumi.ResourceOptions)` check to pulumi/pulumi.

pkg/tfgen/generate_python.go
```diff
@@ -535,7 +529,6 @@ func (g *pythonGenerator) emitResourceType(mod *module, res *resourceType) (stri
 			arg = fmt.Sprintf("pulumi.Output.from_input(%s).apply(json.dumps) if %s is not None else None", arg, arg)
 		}
 		w.Writefmtln("        __props__['%s'] = %s", pname, arg)
-		w.Writefmtln("")
```
CyrusNajmabadi: we have so much blank lines in __init__, just trying to clean things up.

pkg/tfgen/generate_python.go
```diff
@@ -585,8 +578,8 @@ func (g *pythonGenerator) emitResourceType(mod *module, res *resourceType) (stri
 
 	// Override translate_{input|output}_property on each resource to translate between snake case and
 	// camel case when interacting with tfbridge.
-	w.Writefmtln(`
```
pgavlin: Do we not need this newline in the output?

pkg/tfgen/generate_python.go
pgavlin: In order for this to be completely safe, we'll need to bump the minimum version of the Pulumi Python SDK to one that includes the `ResourceOptions` check.

CyrusNajmabadi:
> In order for this to be completely safe, we'll need to bump the minimum version of the Pulumi Python SDK to one that includes the ResourceOptions check.

You're definitely correct.  However, It's only unsafe in that it might allow something to be passed through that wasn't legal before.  I'm personally ok with that.  But if we're really concerned, i can put that specific check back :)

pkg/tfgen/generate_python.go
CyrusNajmabadi: we do not.  There is already a blank line above.  This leads to a slightly ungainly double blank line.

CyrusNajmabadi: It's also not a big deal.  the primary thing i wanted to improve was the awful assignment into props above.  This was just tiny cleanup along the way.

pkg/tfgen/generate_python.go
```diff
@@ -493,12 +493,6 @@ func (g *pythonGenerator) emitResourceType(mod *module, res *resourceType) (stri
 	w.Writefmtln(
 		`            warnings.warn("explicit use of __opts__ is deprecated, use 'opts' instead", DeprecationWarning)`)
 	w.Writefmtln("            opts = __opts__")
```
CyrusNajmabadi:
```suggestion
	w.Writefmtln("            opts = __opts__")
	w.Writefmtln("        if opts and not isinstance(opts, pulumi.ResourceOptions):")
	w.Writefmtln("            raise TypeError('Expected resource options to be a ResourceOptions instance')")
```

PR #451, opened 2019-08-07
Update Terraform to v0.12.6

jen20:
This updates to the latest version of Terraform in order that late versions can be used with `RemoteStateReference`.


jen20:
We'll need to cut a release after this to get the latest version out in the `pulumi-terraform` package.

jen20:
The downstream failure here looks to be unrelated to this change.

PR #452, opened 2019-08-07
Add support for `Resource.get` in python.

CyrusNajmabadi:
Fixes https://github.com/pulumi/pulumi/issues/1123

Example of what this looks like in the emitted code:

```python
    @staticmethod
    def get(resource_name, id, opts=None, certificate_arn=None, validation_record_fqdns=None):
        """
        Get an existing CertificateValidation resource's state with the given name, id, and optional extra
        properties used to qualify the lookup.
        :param str resource_name: The unique name of the resulting resource.
        :param str id: The unique provider ID of the resource to lookup.
        :param pulumi.ResourceOptions opts: Options for the resource.
        :param pulumi.Input[str] certificate_arn: The ARN of the certificate that is being validated.
        :param pulumi.Input[list] validation_record_fqdns: List of FQDNs that implement the validation. Only valid for DNS validation method ACM certificates. If this is set, the resource can implement additional sanity checks and has an explicit dependency on the resource that is implementing the validation

        > This content is derived from https://github.com/terraform-providers/terraform-provider-aws/blob/master/website/docs/r/acm_certificate_validation.html.markdown.
        """
        opts = pulumi.ResourceOptions(id=id) if opts is None else opts.merge(pulumi.ResourceOptions(id=id))

        __props__ = dict()
        __props__["certificate_arn"] = certificate_arn
        __props__["validation_record_fqdns"] = validation_record_fqdns
        return CertificateValidation(resource_name, opts=opts, __props__=__props__)
```


pkg/tfgen/generate_python.go
```diff
@@ -493,13 +495,16 @@ func (g *pythonGenerator) emitResourceType(mod *module, res *resourceType) (stri
 	w.Writefmtln(
 		`            warnings.warn("explicit use of __opts__ is deprecated, use 'opts' instead", DeprecationWarning)`)
 	w.Writefmtln("            opts = __opts__")
-	w.Writefmtln("        if opts and not isinstance(opts, pulumi.ResourceOptions):")
+	w.Writefmtln("        if opts is None:")
+	w.Writefmtln("            opts = pulumi.ResourceOptions()")
```
CyrusNajmabadi: this moved up from below.  it simplifies things so we don't need later null checks.

pkg/tfgen/generate_python.go
```diff
@@ -493,13 +495,16 @@ func (g *pythonGenerator) emitResourceType(mod *module, res *resourceType) (stri
...
+	w.Writefmtln("        if not isinstance(opts, pulumi.ResourceOptions):")
 	w.Writefmtln("            raise TypeError('Expected resource options to be a ResourceOptions instance')")
-	w.Writefmtln("")
-
-	// Now copy all properties to a dictionary, in preparation for passing it to the base function.  Along
-	// the way, we will perform argument validation.
-	w.Writefmtln("        __props__ = dict()")
+	w.Writefmtln("        if opts.version is None:")
+	w.Writefmtln("            opts.version = utilities.get_version()")
```
CyrusNajmabadi: this moved up from below.

pkg/tfgen/generate_python.go
```diff
@@ -989,36 +1003,60 @@ func (g *pythonGenerator) emitInitDocstring(w *tools.GenWriter, mod *module, res
...
+	fmt.Fprintln(&buf, ":param pulumi.ResourceOptions opts: Options for the resource.")
+	for _, prop := range res.outprops {
+		g.emitPropDocstring(&buf, prop)
 	}
 
 	// emitDocComment handles the prefix and triple quotes.
 	g.emitDocComment(w, buf.String(), res.docURL, "        ")
 }
 
+func (g *pythonGenerator) emitPropDocstring(buf io.Writer, prop *variable) {
```
CyrusNajmabadi: extracted code to be used by `emitGetDocstring` and `emitInitDocstring`

CyrusNajmabadi:
Need eyes please :)  @jen20 @stack72 can you ptal?

pkg/tfgen/generate_python.go
```diff
@@ -578,6 +573,24 @@ func (g *pythonGenerator) emitResourceType(mod *module, res *resourceType) (stri
...
 
+	w.Writefmtln("    @staticmethod")
+	w.Writefmt("    def get(resource_name, id, opts=None")
+	for _, prop := range res.outprops {
+		w.Writefmt(", %[1]s=None", pycodegen.PyName(prop.name))
+	}
+	w.Writefmtln("):")
+	g.emitGetDocstring(w, mod, res)
+	w.Writefmtln(
+		"        opts = pulumi.ResourceOptions(id=id) if opts is None else opts.merge(pulumi.ResourceOptions(id=id))")
```
lukehoban: Would love to open an issue to track adding a version of `merge` that does this by default.  It's a shame that we'll have to encourage to write the kind of code themselves to use `merge` correctly.

lukehoban:
Any chance you could share an example of the new `CertificateValidation` constructor that goes along with the `get` method in the example above?

CyrusNajmabadi:
> Any chance you could share an example of the new CertificateValidation constructor that goes along with the get method in the example above?

Sure, it's:

```py
    def __init__(__self__, resource_name, opts=None, certificate_arn=None, validation_record_fqdns=None, __props__=None, __name__=None, __opts__=None):
        """
        ... trimmed
        """
        > This content is derived from https://github.com/terraform-providers/terraform-provider-aws/blob/master/website/docs/r/acm_certificate_validation.html.markdown.
        """
        if __name__ is not None:
            warnings.warn("explicit use of __name__ is deprecated", DeprecationWarning)
            resource_name = __name__
        if __opts__ is not None:
            warnings.warn("explicit use of __opts__ is deprecated, use 'opts' instead", DeprecationWarning)
            opts = __opts__
        if opts is None:
            opts = pulumi.ResourceOptions()
        if not isinstance(opts, pulumi.ResourceOptions):
            raise TypeError('Expected resource options to be a ResourceOptions instance')
        if opts.version is None:
            opts.version = utilities.get_version()
        if opts.id is None:
            if __props__ is not None:
                raise TypeError("[__props__] should only be provided when [opts.id] was not [None].")
            __props__ = dict()

            if certificate_arn is None:
                raise TypeError("Missing required property 'certificate_arn'")
            __props__['certificate_arn'] = certificate_arn
            __props__['validation_record_fqdns'] = validation_record_fqdns
        super(CertificateValidation, __self__).__init__(
            'aws:acm/certificateValidation:CertificateValidation',
            resource_name,
            __props__,
            opts)
```

pkg/tfgen/generate_python.go
```diff
@@ -493,13 +495,16 @@ func (g *pythonGenerator) emitResourceType(mod *module, res *resourceType) (stri
...
-	w.Writefmtln("")
-
-	// Now copy all properties to a dictionary, in preparation for passing it to the base function.  Along
-	// the way, we will perform argument validation.
-	w.Writefmtln("        __props__ = dict()")
+	w.Writefmtln("        if opts.version is None:")
+	w.Writefmtln("            opts.version = utilities.get_version()")
+	w.Writefmtln("        if opts.id is None:")
+	w.Writefmtln("            if __props__ is not None:")
+	w.Writefmtln("                raise TypeError(\"[__props__] should only be provided when [opts.id] was not [None].\")")
```
lukehoban: We don't use this `[...]` style anywhere else in similar error messages. Also - this is difficult for me to parse - perhaps instead?

```
__props__ is only valid when passed in combination with a valid opts.id to get an existing resource
```

CyrusNajmabadi: great.

pkg/tfgen/generate_python.go
CyrusNajmabadi: https://github.com/pulumi/pulumi/issues/3058

pkg/tfgen/generate_python.go
```diff
@@ -481,6 +482,7 @@ func (g *pythonGenerator) emitResourceType(mod *module, res *resourceType) (stri
 
 	// Old versions of TFGen emitted parameters named __name__ and __opts__. In order to preserve backwards
 	// compatibility, we still emit them, but we don't emit documentation for them.
+	w.Writefmt(", __props__=None")
```
pgavlin: Do we actually need `__props__`? I thought we kind of figured out that this was a mistake in Node...

CyrusNajmabadi: High order bit for now: parity with Nodejs.  Decision about removing props from being used in reads is punted to a later point when we can make the decision for both.

PR #453, opened 2019-08-07
Split CHANGELOG for bridge vs provider

jen20:
This PR splits the CHANGELOG files for the bridge and the provider.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,36 +1,21 @@
-# `@pulumi/terraform` CHANGELOG
+# `pulumi-terraform` CHANGELOG
 
 This CHANGELOG details important changes made in each version of the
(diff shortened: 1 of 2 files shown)


lukehoban:
Will this work with our changelog required check?  I believe that will enforce that the main changelog be updated. We may need to update that tool to allow any changelog to be updated?

jen20:
@lukehoban It won't work with it right now, but should be a straightforward update. Until that's done we can use use the `impact/no-changelog-required` label, as this definitely simplifies things for consumers of the provider in terms of understanding what has changed between versions in the user-facing context.

jen20:
This is no longer needed, as splitting out `pulumi-terraform-bridge` should leave the CHANGELOG here covering only the provider.

PR #455, opened 2019-08-08
Generate awaitable subclasses for datasource results.

pgavlin:
The prior approach--deleting the __await__ and __iter__ methods before
returning self--does not work on Python 3.6. This approach seems as if
it will be as backwards-compatible as possible while still allowing us
to return something like a plain-old value for datasource methods.

### DIFF
--- a/pkg/tfgen/generate_python.go
+++ b/pkg/tfgen/generate_python.go
@@ -416,9 +416,11 @@ func (g *pythonGenerator) emitRawDocComment(w *tools.GenWriter, comment, prefix 
 	}
 }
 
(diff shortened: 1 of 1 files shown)


CyrusNajmabadi:
Looks fine to me.  Wouldn't mind an example of how the emit for this looks.

pgavlin:
@CyrusNajmabadi Here's the results of one of the `diff` jobs: https://github.com/pulumi-bot/pulumi-aws/compare/dcabdd89fd0a2a0d7424b5e81227340f8cc63bab...42c1f9659cd3cccec71ccb9e4fb1fa28deb18a37

PR #459, opened 2019-08-15
Resolve setup issues

ibrasho:
I'm trying to use the repo for a new project and faced some issues.

This PR attempts to resolve the following issues:
* Link to the correct contributing guidelines: It seems the repo moved and the links were not updated.
* Replace dep with go mod: dep was removed a couple of months ago, but the build instructions still mention dep.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,6 @@
 
 Do you want to hack on Pulumi?  Awesome!  We are so happy to have you.
 
-Please refer to the [main Pulumi repo](https://github.com/pulumi/home/)'s [CONTRIBUTING.md file](
-https://github.com/pulumi/home/blob/master/CONTRIBUTING.md) for details on how to do so.
+Please refer to the [docs Pulumi repo](https://github.com/pulumi/docs/)'s [CONTRIBUTING.md file](
+https://github.com/pulumi/docs/blob/master/CONTRIBUTING.md) for details on how to do so.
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -14,6 +14,7 @@ PYPI_VERSION     := $(shell scripts/get-py-version)
 VERSION_FLAGS    := -ldflags "-X github.com/pulumi/pulumi-terraform/pkg/version.Version=${VERSION}"
 
 build::
+	GO111MODULE=on go mod vendor
```
lukehoban: We have a separate `make ensure` task that will take care of this step.  We should document that in the README/CONTRIBUTING, but for consistency with the rest of the Pulumi repos, I don't think we want to change what `make build` means for this repo.

CONTRIBUTING.md
```diff
@@ -2,6 +2,6 @@
 
 Do you want to hack on Pulumi?  Awesome!  We are so happy to have you.
 
-Please refer to the [main Pulumi repo](https://github.com/pulumi/home/)'s [CONTRIBUTING.md file](
-https://github.com/pulumi/home/blob/master/CONTRIBUTING.md) for details on how to do so.
+Please refer to the [docs Pulumi repo](https://github.com/pulumi/docs/)'s [CONTRIBUTING.md file](
+https://github.com/pulumi/docs/blob/master/CONTRIBUTING.md) for details on how to do so.
```
lukehoban: @ellismg Do we need to fix this up more broadly?

jen20: I think this is still needed - I'll cherry pick it on top of the latest bits.

jen20:
Hi @ibrasho! The change to Go modules has already happened, but I think the link update is still valid - I'll cherry pick that piece of your change onto the latest branches. I'll go ahead and close out this PR in the meantime. Thanks for opening a PR here!

PR #462, opened 2019-08-15
Ignore state-only diffs.

pgavlin:
It is possible for a Terraform diff to contain diffs to properties that
are only present in the state. These diffs typically exist because
`CustomizeDiff` has marked a computed property as differing (see e.g.
the `last_modified` property of `aws_lambda_function`). We do not care
about these diffs and can safely remove them from the detailed diff.


pgavlin:
(related to https://github.com/pulumi/pulumi-service/issues/3774)

PR #463, opened 2019-08-15
Reflect assets from inputs to outputs.

pgavlin:
If an input property value `p` is an asset or archive, reflect the same
asset or archive into the output property `p` if it exists.

This fixes spurious diffs due to differences between the temporary
paths we use for spilling assets when a stack is updated across
machines/reboots/etc.


pgavlin:
(related to https://github.com/pulumi/pulumi-service/issues/3774)

PR #465, opened 2019-08-16
Set delete-before-replace for manually-named resources.

### LINKED ISSUE #383: Consider setting `deleteBeforeReplace` by default when a resource is not autonamed
Today, we default to creating before delete on replacement.  Users can manually provide `deleteBeforeRepalce: true` to force the delete to happen first.

For most resources, if the `name: ` property is specified explicitly, then it is not possible to create before delete, as the names will conflict.  If autonaming is applied (the default), then this will not be a problem.

We should consider coupling these two more closely - making `deleteBeforeReplace: true` for resources that are not autonamed.

pgavlin:
Determine if a resource is auto-named by checking its list of top-level
defaults for a property that was populated by a default value that is an
autoname. If the defaults list is missing, consider the resource
auto-named for better backwards compatibility.

These changes also fix a bug in the logic that determined the legacy
replaces list.

Fixes #383.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ This CHANGELOG details important changes made in each version of the
 - The Python SDK generated for a provider now supports calling `SomeResource.get(...)` to create a
   resource with the state of existing cloud resource.
 - Generate named nested types with doc comments instead of anonymous inline expansions for TypeScript.
+- Set `DeleteBeforeReplace` for resources that are not auto-named.
 
 ## v0.18.3 (Released June 20, 2019)
 
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -608,8 +608,8 @@ func (p *Provider) Diff(ctx context.Context, req *pulumirpc.DiffRequest) (*pulum
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


pgavlin:
> Would love to verify that deleteBeforeReplace: false will work as a way to override this if needed.

As of this writing, that will not work. We will need to change the CLI in order to provide that override.

pgavlin:
One thing I should point out: this will enable DBR for any resource that is not autonamed, which is subtly different than enabling DBR only for resources that are manually-named. With these changes, any resource that does not have an autonameable property will always be DBR'd. There are 276 such resources in the AWS provider at the moment; I haven't looked at the others.

pgavlin:
I have added a commit that changes the way we decide whether or not a resource's name requires DBR: rather than looking for the presence of a default that came from an autoname, the code now looks for an input property with a manually-specified (i.e. non-default) value that _could_ have been autonamed.

pgavlin:
I have manually validated that this works as expected with resources that are not auto-nameable (e.g. EC2 instances) as well as with resources that are. I have also validated that `deleteBeforeReplace: false` works as expected when using the latest dev versions of the CLI and SDK.

PR #471, opened 2019-08-21
Allow IDs to change during import.

pgavlin:
It is possible for a resource's import function to change the ID of the
resource. The current code does not handle this, leading to resources
that cannot be read or imported.

Fixes #708.


pgavlin:
> I believe this solves a lot of import problems where we have IDs like X/y e.g. RouteTableAssociations - correct?

This needs a small change in the CLI as well, but yes.

PR #472, opened 2019-08-21
Accept Archives everywhere Assets are accepted.

pgavlin:
Just what it says on the tin. If no format has been specified in the
asset translation, default to ZIP.

Fixes https://github.com/pulumi/pulumi/issues/280.

### DIFF
--- a/pkg/tfbridge/assets.go
+++ b/pkg/tfbridge/assets.go
@@ -156,22 +156,24 @@ func (a *AssetTranslation) TranslateAsset(asset *resource.Asset) (interface{}, e
 
 // TranslateArchive translates the given archive using the directives provided by the translation info.
 func (a *AssetTranslation) TranslateArchive(archive *resource.Archive) (interface{}, error) {
-	contract.Assert(a.IsArchive())
-
 	// TODO[pulumi/pulumi#153]: support HashField.
 
 	// Produce either a temp file or an in-memory representation, as requested.
+	format := a.Format
+	if format == resource.NotArchive {
+		format = resource.ZIPArchive
+	}
 	switch a.Kind {
-	case FileArchive:
+	case FileArchive, FileAsset:
 		path, err := translateToFile(archive.Hash, archive.HasContents(), func(w io.Writer) error {
-			return archive.Archive(a.Format, w)
+			return archive.Archive(format, w)
 		})
 		return path, err
-	case BytesArchive:
+	case BytesArchive, BytesAsset:
 		if !archive.HasContents() {
 			return []byte{}, nil
 		}
-		return archive.Bytes(a.Format)
+		return archive.Bytes(format)
 	default:
 		contract.Failf("Unrecognized asset translation kind: %v", a.Kind)
 		return nil, nil
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ This CHANGELOG details important changes made in each version of the
   resource with the state of existing cloud resource.
 - Generate named nested types with doc comments instead of anonymous inline expansions for TypeScript.
 - Set `DeleteBeforeReplace` for resources that are not auto-named.
+- Accept archive values everywhere asset values are accepted.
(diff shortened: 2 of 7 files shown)


pgavlin:
I have confirmed that this works using the following programs:

```typescript
import * as pulumi from "@pulumi/pulumi";
import * as aws from "@pulumi/aws";

// Create an AWS resource (S3 Bucket)
const bucket = new aws.s3.Bucket("my-bucket");

// Create a bucket object with the contents of a folder.
const obj = new aws.s3.BucketObject("archive", {
    bucket: bucket.id,
    source: new pulumi.asset.FileArchive("./content"),
});

// Export the name of the bucket
export const bucketName = bucket.id;
```

```python
import pulumi
from pulumi_aws import s3

# Create an AWS resource (S3 Bucket)
bucket = s3.Bucket('my-bucket')

obj = s3.BucketObject('archive', bucket=bucket.id, source=pulumi.FileArchive('./content'))

# Export the name of the bucket
pulumi.export('bucket_name',  bucket.id)
```

pkg/tfbridge/assets.go
```diff
@@ -156,22 +156,24 @@ func (a *AssetTranslation) TranslateAsset(asset *resource.Asset) (interface{}, e
...
 func (a *AssetTranslation) TranslateArchive(archive *resource.Archive) (interface{}, error) {
-	contract.Assert(a.IsArchive())
-
 	// TODO[pulumi/pulumi#153]: support HashField.
 
 	// Produce either a temp file or an in-memory representation, as requested.
+	format := a.Format
+	if format == resource.NotArchive {
+		format = resource.ZIPArchive
+	}
```
lukehoban: Do we want to make the `ResourceInfo` author specify the desired archive kind to opt into this?  That at least ensures that this can be added incrementally without breaking changes.

pgavlin: I think that accepting both an `Asset` and an `Archive` is the correct default behavior. I worry that this is subtle enough that schema authors won't know that they should set a value for `Format` when choosing an `Asset`-typed translation. I also don't think that changing the archive format is necessarily a breaking change.

pgavlin:
> I assume for cases 2 and 3 we can leave those as-is. They indicate that only and Archive is allowed. And you can use FileArchive to construct an Archive from an arbitrary file on disk. That is - if the cloud API expects something that is a ZIP or TGZ or whatever, we will keep using either (2) or (3). This also is the only way that HashField can be used (is it true that that is only useful when you know the thing is an archive? that isn't obvious).

Correct.

> For case 1, we generally believe that anything that indicates it accepts an Asset could also handle being passed an Archive, and so we implicitly accept either here. Any place where an arbitrary file is expected should be mapped as tfbridge.FileAsset and will get this by default, and will default to ZIP, but that could be overridden by the author of the mapping.

Correct.

> In practice, we will change a couple of places that currently incorrectly use (2) above to use (1), which will not be a breaking change.

That's the plan, yes.

joeduffy:
This is a very cool approach, one that I hadn't even contemplated!

Do we think this replaces the need to somehow enable an `ArchiveAsset` class in the Pulumi SDK, that would take any archive and make an asset out of it?

Or is this just a stop gap that accomplishes most of the benefit while also minimizing the scope of adding such a feature -- but that we will eventually want the lower level compositional capabilities?

pgavlin:
> Do we think this replaces the need to somehow enable an ArchiveAsset class in the Pulumi SDK, that would take any archive and make an asset out of it?

I do think that this replaces that need, yes. We may be able to do some work in `pulumi/pulumi/pkg/resource` to ease the burden on provider authors in the future, but I don't think we need another public-facing type in the SDK.

pkg/tfbridge/assets.go
lukehoban: After working through the examples of how we'll use this in practice - I agree.  If we hit a case where we need more flexibility - we can extend as needed.

PR #483, opened 2019-08-30
Upgrade hashicorp/terraform dependency

stack72:
The Terraform codebase has moved a package since 0.12.7. Any providers
that have upgraded their dependency on Terraform will now fail due to
this package moving

This is not something we are going to be able to roll out to all
providers. It will need to be rolled out provider by provider or the
providers will not build

@pgavlin we need to think about this - right now, only the cloudflare provider is broken because that's the first provider with this update. I would almost suggest we wait for a week to merge it to stop it blocking any of the urgent fixes we need to get in

Thoughts?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,10 +11,9 @@ require (
 	github.com/golang/protobuf v1.3.1
 	github.com/hashicorp/go-multierror v1.0.0
 	github.com/hashicorp/hcl v0.0.0-20170504190234-a4b07c25de5f
-	github.com/hashicorp/terraform v0.12.6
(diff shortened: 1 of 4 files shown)


pgavlin:
> right now, only the cloudflare provider is broken because that's the first provider with this update

What happens if we try to build the cloudflare provider with `terraform` overridden to 0.12.6?

stack72:
@pgavlin we get an error that the hcl2shim package isn't found

stack72:
@pgavlin I expect the downstream deps to fail with this :( Until they are upgraded to the latest version of Terraform

PR #494, opened 2019-09-14
Remove the need for `pandoc` during the build

stack72:
Continues the work from https://github.com/pulumi/pulumi/pull/3195
for the pulumi-providers based on TF

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ This CHANGELOG details important changes made in each version of the
 - Include nested structure details in Python docstrings.
 - Use `pulumi.InvokeOptions()` when `opts` is `None` for Python data source functions.
 - Provide a mechanism for overriding nested type names.
+- Remove the need for Pandoc in generating Python SDK readme files.
 
 ## v0.18.3 (Released June 20, 2019)
 
--- a/pkg/tfgen/generate_python.go
+++ b/pkg/tfgen/generate_python.go
@@ -831,7 +831,7 @@ func (g *pythonGenerator) emitPackageMetadata(pack *pkg) error {
 	// Generate a readme method which will load README.rst, we use this to fill out the
 	// long_description field in the setup call.
 	w.Writefmtln("def readme():")
-	w.Writefmtln("    with open('README.rst') as f:")
+	w.Writefmtln("    with open('README.md', encoding='utf-8') as f:")
 	w.Writefmtln("        return f.read()")
 	w.Writefmtln("")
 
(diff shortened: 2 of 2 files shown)


pgavlin:
The "test codegen" are all failing due to this change. I assume we'll need to adjust something in the providers themselves (or the common scripts repo) after checking this in and pushing it out?

stack72:
I expect Azure to fail due to it not being on Terraform 0.12.7

The other 2 providers are due to go mod tidy not being run - I am confident this is ok

pgavlin:
> The other 2 providers are due to go mod tidy not being run - I am confident this is ok

Are you sure that's the case? From [the logs](https://travis-ci.com/pulumi/pulumi-terraform/jobs/235316109#L1948-L1962):
```
cd sdk/python/ && \
	if [ $(command -v pandoc) ]; then \
		pandoc --from=markdown-smart --to=rst-smart --output=README.rst ../../README.md; \
	else \
		echo "warning: pandoc not found, not generating README.rst"; \
		echo "" > README.rst; \
	fi && \
	python3 setup.py clean --all 2>/dev/null && \
	rm -rf ./bin/ ../python.bin/ && cp -R . ../python.bin && mv ../python.bin ./bin && \
	sed -i.bak -e "s/\${VERSION}/1.2.2a1568461668/g" -e "s/\${PLUGIN_VERSION}/v1.2.2-alpha.1568461668+g7fc04bf4/g" ./bin/setup.py && \
	rm ./bin/setup.py.bak && \
	cd ./bin && python3 setup.py build sdist
Makefile:26: recipe for target 'build' failed
make: *** [build] Error 1
```

The `build` target does not depend on the `ensure` target, so I'm not sure why a lack of `go mod tidy` would cause it to fail. It looks like the problem is the change to the `setup.py` script, as the downstream repos don't have a `README.md` to open.

stack72:
@pgavlin my apologies here, I must have been in a world of my own when I wrote that comment :D

There will be a change needed to be made to all of the providers to take advantage of this. So we will need to change their Makefile to look as follows:


before
```
if [ $$(command -v pandoc) ]; then \
			pandoc --from=markdown-smart --to=rst-smart --output=README.rst ../../README.md; \
		else \
			echo "warning: pandoc not found, not generating README.rst"; \
			echo "" > README.rst; \
		fi && \
```

after
```
cp ../../README.md $(PYENVSRC)
```

So I do expect the downstream builds to fail based on this test. FWIW, this will close:

https://github.com/pulumi/pulumi-azure/issues/370
and supercede https://github.com/pulumi/pulumi-gcp/pull/74

jen20:
This is now superseded by #500, which includes the correct fix to remove the `pandoc` requirement.

PR #496, opened 2019-09-20
Fix raw element names in detailed diff.

pgavlin:
The code that computes a detailed diff was not properly handling maps
that require raw element names.

Fixes https://github.com/pulumi/pulumi-aws/issues/751.


pgavlin:
> Is there any higher level integration testing that could/should have caught this?

We could have caught this with a test that changes an AWS Lambda envvar.

pgavlin:
CI failure is an expected Azure job failure.

PR #498, opened 2019-09-26
Make tfbridge and tfgen use the new `terraform-plugin-sdk`

jen20:
This ports the `tfbridge` and `tfgen` components to use the new `github.com/hashicorp/terraform-plugin-sdk`. I _do not_ think this should be merged into this repo, but should instead move to a new repository named `pulumi-terraform-bridge`, which other providers are then updated to depend on, leaving this repo free for the `pulumi-terraform` _provider_ - which necessarily does depend on the full core of Terraform.

One of the tests suggests some behavioural differences around empty defaults - @stack72 suggested this had already been observed when attempted to pull in the latest version of Terraform. @pgavlin, any ideas there (see the diff in `schema_test.go`).

Once we've reviewed and determined the best repo layout for these bits, we can determine the best course of action for this draft PR.


jen20:
I've moved this to the new repository at `pulumi/pulumi-terraform-bridge` for further work.

PR #499, opened 2019-09-26
Do not fail to import resources with no Importer.

pgavlin:
Just what it says on the tin.


pgavlin:
As a heads up, there is _some_ risk in this change: `Read` seems to assume in many cases that the ID is well-formed, so malformed IDs can cause panics in the underlying Terraform provider.

PR #500, opened 2019-09-30
Remove Terraform Bridge components

jen20:
Since the Terraform Plugin SDK was released, the Terraform Bridge has no need to depend on Terraform itself - however the resource provider which provides `RemoteStateReference` does still need the full dependency. In order to avoid imposing this on providers, the bridge code (without the provider) now lives in the `pulumi-terraform-bridge` repository instead, so this commit removes the duplicate code.

This likely does not want merging until `pulumi-terraform-bridge` is made public.


sdk/python/setup.py
```diff
@@ -22,14 +22,15 @@ def run(self):
...
-    with open('README.rst') as f:
+    with open('README.md') as f:
         return f.read()
 
 
 setup(name='pulumi_terraform',
       version='${VERSION}',
       description='A Pulumi package for consuming Terraform Remote State resources.',
       long_description=readme(),
+      long_description_content_type='test/markdown',
```
pgavlin: Should this be `text/markdown`?

stack72: Yes it should

jen20: Fixed in d3a1bd1.

PR #501, opened 2019-09-30
Remove our dependency on flatmap.

pgavlin:
These changes remove our dependency on the flatmap package. The
functionality we needed from that package is trivial to replace with a
bit of straighforward code.

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -20,12 +20,12 @@ import (
 	"reflect"
 	"sort"
 	"strconv"
+	"strings"
 
 	"github.com/golang/glog"
 
 	pbstruct "github.com/golang/protobuf/ptypes/struct"
 	"github.com/hashicorp/terraform/configs/hcl2shim"
-	"github.com/hashicorp/terraform/flatmap"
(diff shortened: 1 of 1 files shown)


pkg/tfbridge/schema.go
```diff
@@ -501,7 +501,22 @@ func MakeTerraformResult(state *terraform.InstanceState,
...
 			Map:    schema.BasicMapReader(attrs),
 		}
-		for _, key := range flatmap.Map(attrs).Keys() {
+
+		// Read each top-level field out of the attributes.
+		keys := make(map[string]bool)
+		for key := range attrs {
+			// Pull the top-level field out of this attribute key. If we've already read the top-level field, skip this
+			// key.
+			dot := strings.Index(key, ".")
```
lukehoban: Is there only ever one `.` in these attrs?

pgavlin: No: there is frequently more than one `.`. We are specifically looking for the _first_ dot, as we only need the component of the attribute key that precedes that first dot. That portion of the key is the name of the top-level property in which attribute is contained.

pgavlin:
Ran the AWS tests locally with these changes; everything looks good.

PR #508, opened 2019-10-26
Inherit child options args from ResourceArgs

mikhailshilkov:
Required for serialization, e.g.
```cs
    public class VirtualNetworkSubnetsArgs : Pulumi.ResourceArgs
```


mikhailshilkov:
@pgavlin

PR #531, opened 2020-05-01
Update to Terraform v0.12.24 and restructure for v2

jen20:
This PR updates `pulumi-terraform` to work with state files from Terraform v0.12.24 and below. We also restructure to use the new Pulumi v2 modules.

Thanks to @jaxxstorm for rubber-ducking various parts of restructuring!

Still required is a test of reading the new state format for Terraform v0.12.24, since all of our current examples have state generated by v0.11.


jen20:
Interesting, the linter running out of memory was not the expected failure mode of this...

jen20:
I guess we need to bump the version of the linter to have it pass successfully here since all is well with it disabled.

.travis.yml
```diff
@@ -1,7 +1,7 @@
 # It may be tempting to add parens around each individual clause in this expression, but Travis then builds pushes anyway
 if: branch = master OR branch =~ ^release/ OR branch =~ ^features/ OR tag IS present
 language: go
-go: 1.12.x
+go: 1.14.x
```
stack72, line 4: FWIW, nothing else is upgraded to v1.14.x yet... that will start next week :)

scripts/publish-plugin.sh
```diff
@@ -17,9 +17,9 @@ if [ "$(go env GOOS)" = "windows" ]; then
 fi
 
 go build \
-   -ldflags "-X github.com/pulumi/pulumi-terraform/pkg/version.Version=${VERSION}" \
+   -ldflags "-X github.com/pulumi/pulumi-terraform/provider/cmd/pulumi-resource-terraform/main.Version=${VERSION}" \
```
stack72, line 20: I am pretty sure this needs to be

```
(cd "${ROOT}/provider" && go build \
```

.travis.yml
jaxxstorm: I'll switch this back to `1.13.x` in the next push

jen20: 1.13 appears to exhibit different behaviour with regards to building modules from outside of their directory (or at least I thought that's what it did last night) - if we're upgrading to 1.14 this week anyway it might be better to use this as a test bed?

scripts/publish-plugin.sh
jen20: Fixed in e2cfb53.

PR #540, opened 2020-05-18
Adding examples for RemoteBackend

stack72:
Fixes: #539


sdk/go/state/internal/remote_args.go
```diff
@@ -37,3 +37,7 @@ type workspaceArgs struct {
 func (RemoteBackendStateReferenceArgs) ElementType() reflect.Type {
 	return reflect.TypeOf((*remoteBackendStateReferenceArgs)(nil)).Elem()
 }
+
+func (WorkspaceArgs) ElementType() reflect.Type {
```
stack72, line 41: This fixes a small bug in the reflected type that allows us to create a fake generic implementation of the RemoteBackend args

Without this, we were getting a bug that says it could not reflect internal.WorkspaceArgs to internal.workspaceArgs

PR #544, opened 2020-06-03
Use the pulumi-ci AWS profile for CI

praneetloke:
Related to pulumi/home#845.

The `master` branch build failed during the integration test phase due to the AWS creds that were changed with #543 having very limited AWS access. This PR updates the Travis config file to use the CI AWS profile. But that's only one part of it. The other part is that I will update the AWS creds in the Travis settings for this repo to use the corresponding creds that would allow it to use that CI profile.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,10 @@ before_install:
 install:
     - source ${PULUMI_SCRIPTS}/ci/install-common-toolchain.sh
     - source ${PULUMI_SCRIPTS}/ci/configure-aws.sh
+    - export AWS_SECURITY_TOKEN=${AWS_SESSION_TOKEN}
     # Use the pulumi-ci AWS profile configured by the above script.
-    # This repo runs tests that require access to create resources in AWS.
+    # This repo runs tests that require access to resources in AWS besides just the
+    # get.pulumi.com and rel.pulumi.com S3 buckets.
     - export AWS_PROFILE=pulumi-ci
     # Install Pulumi
     - curl -L https://get.pulumi.com/ | bash
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,8 @@ before_install:
 install:
     - source ${PULUMI_SCRIPTS}/ci/install-common-toolchain.sh
     - source ${PULUMI_SCRIPTS}/ci/configure-aws.sh
-    - export AWS_SECURITY_TOKEN=${AWS_SESSION_TOKEN}
     # Use the pulumi-ci AWS profile configured by the above script.
-    # This repo runs tests that require access to resources in AWS besides just the
-    # get.pulumi.com and rel.pulumi.com S3 buckets.
+    # This repo runs tests that require access to create resources in AWS.
     - export AWS_PROFILE=pulumi-ci
     # Install Pulumi
     - curl -L https://get.pulumi.com/ | bash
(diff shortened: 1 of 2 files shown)


praneetloke:
I will update the AWS creds once the PR job completes.

praneetloke:
@chrsmith the integration tests are now unable to detect AWS credentials at all. Previously they were failing because they could not access the S3 buckets referenced in the tests, so they did detect the creds before. But now with this PR, the only addition is to use the CI AWS profile, as well as change the AWS creds to the corresponding creds that can assume that role. I was wondering if you have seen this problem with the other repos?

chrsmith:
@praneetloke this PR looks right. The one missing step might be to set `AWS_SDK_LOAD_CONFIG=1` which forces the use of the AWS configuration file (`~/.aws/config`) rather than ambient AWS creds. We needed to set that once or twice in other repos.

But to recap:

```
# This script will pull the AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY env vars
# and write them to ~/.aws/credentials, and create ~/.aws/config that defines the
# pulumi-ci AWS profile. (So the IAM User can assume role into the `pulumi-ci` account.
 - source ${PULUMI_SCRIPTS}/ci/configure-aws.sh

# ... and the comment you added here is 100% spot on.
# Use the pulumi-ci AWS profile configured by the above script.
# This repo runs tests that require access to create resources in AWS.
 - export AWS_PROFILE=pulumi-ci

praneetloke:
> The one missing step might be to set AWS_SDK_LOAD_CONFIG=1 which forces the use of the AWS configuration file (~/.aws/config) rather than ambient AWS creds. We needed to set that once or twice in other repos.

Ok I updated this PR to give that a shot.

praneetloke:
Thanks for fixing this up, @chrsmith. I'll go ahead and merge this now since the build has passed. I'd like to restore the `master` branch to working condition while I like into why setting `AWS_SDK_LOAD_CONFIG=1` was needed for this repo.

praneetloke:
Alright. I think I know why `AWS_PROFILE` wasn't being honored unless `AWS_SDK_LOAD_CONFIG=1` was set. The AWS Go SDK docs state this:

> By default the SDK will only load the shared credentials file's (~/.aws/credentials) credentials values, and all other config is provided by the environment variables, SDK defaults, and user provided aws.Config values. 

The key point here is "user provided" `aws.Config` values. I think what is happening is that these remote state tests that use S3 backend do not supply a value for the `Profile` property. So the SDK does not use the config file to load the profile config and fails to find the credentials for the profile specified through `AWS_PROFILE`.

Furthermore...

> If the AWS_SDK_LOAD_CONFIG environment variable is set, or SharedConfigEnable option is used to create the Session the full shared config values will be loaded. This includes credentials, region, and support for assume role. In addition the Session will load its configuration from both the shared config file (~/.aws/config) and shared credentials file (~/.aws/credentials). Both files have the same format.

So setting `AWS_SDK_LOAD_CONFIG` to a truthy value causes the SDK to load all shared config files (i.e. both `~/.aws/credentials` as well as `~/.aws/config`) and even though the remote state test does not supply a profile arg, the AWS SDK loads all relevant config files.

That's the best explanation I could come up with based on what I am reading. I _think_ it makes sense to me now. Let me know what you think, @chrsmith.

PR #572, opened 2020-11-17
Bump versions of GHA that still reference add-path

stack72:


stack72:
Self merging as this was largely mechanical and is required for further builds to be successful

PR #606, opened 2021-05-20
Migrate repository to use command-dispatch workflow

stack72:


stack72:
Self merging as this is part of the work to move to command dispatch

PR #617, opened 2021-08-19
Upgrade to Terraform v1.0.5

stack72:
Fixes: #605 
Fixes: #608 
Fixes: #609 
Fixes: #610
Fixes: #611 
Fixes: #612 
Fixes: #615
Fixes: #618


provider/shim/go.mod
```diff
@@ -0,0 +1,18 @@
...
+	github.com/golang/protobuf v1.4.3
+	github.com/hashicorp/terraform v1.0.5
+	github.com/hashicorp/terraform-svchost v0.0.0-20200729002733-f050f53b9734
+	github.com/pkg/errors v0.9.1
+	github.com/pulumi/pulumi/sdk/v3 v3.10.3
+	github.com/stretchr/testify v1.6.1
+	github.com/zclconf/go-cty v1.9.1
+)
+
+replace (
```
t0yv0, line 15: Interesting, this needs to be in replace and not require?

t0yv0 (approved):
Really poor familiarity with this code, but at a glance looks good!

provider/shim/go.mod
stack72: we are replacing the dependency that this is pulling in as both (TF and Pulumi) pull different versions of grpc so we need to the TF version

PR #729, opened 2023-05-18
Upgrade pulumi-terraform to v1.4.6

### LINKED ISSUE #726: Upgrade terraform to v1.4.6
Release details: https://github.com/hashicorp/terraform/releases/tag/v1.4.6

aq17:
Fixes #726


t0yv0:
Do we know what's up with failing tests? Thanks.

aq17:
> Do we know what's up with failing tests? Thanks.

Refreshed/ attempting again 🤞🏼 , don't see anything that differs from previous update PR's

t0yv0:
```
  error: error constructing backend state manager: Remote workspace Terraform version "1.4.5" does not match local Terraform version "1.4.6"
  ```

Perhaps skip this test for now, file an issue and we can proceed with it skipped. There seem to be a few other tests so that's good.

File: examples/examples_dotnet_test.go [hunk 1]
```diff
@@ -90,21 +90,21 @@ func TestDotNetOss012(t *testing.T) {
...
-				"workspaceName": "dev",
-			},
-			Secrets: map[string]string{
-				"tfeToken": getRemoteBackendToken(t),
-			},
-		})
-
-	integration.ProgramTest(t, &test)
-}
+// func TestDotNetRemoteBackend(t *testing.T) {
```
t0yv0, line 93: There's `t.Skip("TODO[pulum/pulumi-terraform#12345]")` form that's better as it's more explicit that there's a test that's currently not working.

aq17:
@t0yv0 interesting, so the test is failing for python now (not just dotnet). Okay to skip it in all languages for now, or should we investigate further?

t0yv0:
Skip everywhere - we should investigate sometime but on the surface

```
  error: error constructing backend state manager: Remote workspace Terraform version "1.4.5" does not match local Terraform version "1.4.6"  
```

Makes it look like the issue is with our test environment setup and not with the provider, not an issue that's affecting users, therefore prioritizing unblocking the upgrade.

t0yv0 (approved):

PR #740, opened 2023-07-28
Don't dispatch to pulumi/registry on release

cnunciato:
Since these calls [always fail](https://github.com/pulumi/registry/actions/runs/5694678322/job/15436318005) (and now create `p1` issues for us), we should probably remove them.

We have a handful of other repos that do the same -- pulumi/pulumi-cloud, pulumi/pulumi-archive, others -- so if this makes sense, let me know and I'll happily submit fixes as I notice them.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -98,19 +98,6 @@ jobs:
         - 14.x
         pythonversion:
         - "3.7"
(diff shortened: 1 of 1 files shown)


aq17 (approved):
Makes sense 
cc @mikhailshilkov

PR #757, opened 2024-03-28
Export PostgresRemoteStateReferenceArgs

julienp:
While working on revamping docs generation using plain typedoc (this is really for the pulumi/pulumi and pulumi/policy nodejs sdks, but some providers like this also use it instead of the newer docs gen), I noticed that `PostgresRemoteStateReferenceArgs` was missing from the output.

I set node to v18 for CI, since older versions are no longer supported. Let me know if you prefer I split that into a separate PR.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -86,18 +86,18 @@ jobs:
       fail-fast: true
       matrix:
         dotnetversion:
-        - 3.1.301
+        - 6.x
         goversion:
-        - 1.18.x
+        - 1.21.x
         language:
         - nodejs
         - python
         - dotnet
         - go
         nodeversion:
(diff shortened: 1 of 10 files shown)


.github/workflows/master.yml
```diff
@@ -95,7 +95,7 @@ jobs:
         - dotnet
         - go
         nodeversion:
-        - 14.x
+        - 18.x
```
julienp, line 98: Versions below 18 are no longer supported by upstream https://nodejs.org/en/about/previous-releases or pu/pu

iwahbe (approved):

julienp:
CI is failing on C# now, updating all language runtimes and not just nodejs...

PR #761, opened 2024-06-27
Unblock tests by pinning to a version of Pulumi

iwahbe:
Fixes https://github.com/pulumi/pulumi-terraform/issues/763

We are pinning to v3.108.1 because it was approximately the latest release when this repo was last updated.

Prerequisite for #760 & #762


thomas11 (approved):
Git history obviously has the background, but a comment on why we're pinning would have been nice.

PR #923, opened 2025-12-30
chore: upgrade go-provider to v1

pgavlin:
This upgrade required changing the shape of the two state reference
invokes to conform with the expecations of go-provider@v1. This should
not cause any user-visible changes.


blampe (approved):

blampe:
I suspect this will fix #900.

PR #925, opened 2025-12-30
chore: update pulumi/pulumi to v3.214.0

pgavlin:
Just what it says on the tin. Doing this manually in lieu of the
automated upgrade process as that seems to be busted at the moment.


guineveresaenger (approved):
that's a large bump! godspeed 🙏

Issue #627, opened 2021-10-28
Make ignoreVersionConflict configurable in remote backend


cedrickring:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Currently it's not possible to override [this setting](https://github.com/hashicorp/terraform/blob/df578afd7e34ae32345d86840edd05b4c05559e8/internal/backend/remote/backend.go#L90) of the remote backend. As a result, the version of the tfe workspace/server must [exactly match](https://github.com/hashicorp/terraform/blob/df578afd7e34ae32345d86840edd05b4c05559e8/internal/backend/remote/backend.go#L647) the bundled (I hope thats the correct term) terraform version. Unfortunately, this makes upgrading the package impossible without upgrading the tfe server too. 

Making this configurable, would allow (dangerously) ignoring the version mismatch and thus would allow using newer version of this library with older tfe server versions. As this library only intends to _read_ the remote state, I don't think it's dangerous to set that flag.

### Affected area/feature

This would only affect the reading of remote state references (as far as I can see).

I'd be happy to provide a PR for that, however I don't know yet if this would be an intended feature by the Pulumi devs.

PR #1016, opened 2026-04-30
fix: getRemoteReference with workspaces.name and version mismatch

### LINKED ISSUE #1015: getRemoteReference with workspaces.name always fails with ErrWorkspacesNotSupported
## Bug

`getRemoteReference` with `workspaces.name` set always fails with:

```
error constructing backend state manager: workspaces not supported
```

## Root Cause

In `provider/state_reference/remote.go:83`, the `Invoke` method passes the actual workspace name to `shim.StateReferenceRead`:

```go
results, err := shim.StateReferenceRead(ctx, "remote", stringOrZero(args.Workspaces.Name), ...)
```

The HashiCorp `remote` backend's `StateMgr` method (in `internal/backend/remote/backend.go:639-644`) expects `"default"` when `workspaces.name` is configured:

```go
func (b *Remote) StateMgr(name string) (statemgr.Full, error) {
    if b.workspace == "" && name == backend.DefaultStateName {
        return nil, backend.ErrDefaultWorkspaceNotSupported
    }
    if b.prefix == "" && name != backend.DefaultStateName {
        return nil, backend.ErrWorkspacesNotSupported  // <-- always hits this
    }
    // ...
    case name == backend.DefaultStateName:
        name = b.workspace  // uses the configured workspace name
```

When `workspaces.name` is set, `b.prefix` is empty. If `name` is anything other than `"default"`, it returns `ErrWorkspacesNotSupported`. The backend reads the actual workspace name from its own config when `StateMgr("default")` is called.

## Fix

Pass `"default"` to `StateMgr` when `workspaces.name` is set. The workspace name is already in the backend config.

jdavredbeard:
## Summary

- Fix #1015
- Skip version conflict check for read-only state operations (#627)
- Extract named `VersionConflictIgnorer` interface (review feedback)
- Add integration test exercising both fixes end-to-end

### workspaces.name (#1015)

`remote.go` passed the actual workspace name to the HashiCorp `remote` backend's `StateMgr`. When `workspaces.name` is configured, the backend expects the sentinel value `"default"` — it reads the actual workspace name from its own config. Passing anything else triggers `ErrWorkspacesNotSupported`.

**Fix:** Pass `"default"` when `workspaces.name` is set. Extracted the logic into `Workspaces.stateMgrName()`.

### Version mismatch (#627)

The `remote` backend checks that the local Terraform version matches the workspace's configured version. Since `pulumi-terraform` only reads state, this check is unnecessary and prevents reading from workspaces with a different TF version than the embedded 1.5.5.

**Fix:** Call `IgnoreVersionConflict()` on backends that support it before calling `StateMgr`. Extracted the interface into a named `VersionConflictIgnorer` type shared between the shim and tests.

## Test Plan

- [x] Unit tests for `Workspaces.stateMgrName()` — 4 cases covering name, prefix, both nil, and name precedence
- [x] Unit test for `StateReferenceRead` with local backend
- [x] Unit test confirming remote backend implements `VersionConflictIgnorer`
- [x] Integration test: creates a temporary TFC workspace (TF 1.9.8, mismatched from embedded 1.5.5), uploads state with known outputs, runs a Pulumi program using `getRemoteReference` with `workspaces.name`, and asserts the outputs match — exercises both fixes end-to-end
- [x] Integration test passes in CI with manual workflow patch: https://github.com/pulumi/pulumi-terraform/actions/runs/25390722167/job/74465106560

> **Note:** Integration tests require `TFE_ORGANIZATION` and `TFE_TOKEN` env vars in the test step, which are currently missing from the ci-mgmt-generated workflow. Tracked in #1021.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


provider/shim/shim.go
```diff
@@ -49,6 +55,13 @@ func StateReferenceRead(
 			diagnostics.ErrWithWarnings())
 	}
 
+	// Since we only read state, skip the version conflict check that requires
+	// the local Terraform version to match the remote workspace's version.
+	// Fixes https://github.com/pulumi/pulumi-terraform/issues/627
+	if b, ok := backend.(interface{ IgnoreVersionConflict() }); ok {
```
guineveresaenger: nit: can we name this interface, like we do in the test, and then reference it from the test as well?

guineveresaenger (changes requested):

guineveresaenger:
this looks good to me otherwise! :)

guineveresaenger (approved):
in general, we don't really need the multi-language integration test pattern anymore, since the engine's language conformance testing is now very reliable, so the YAML test would be enough. But I don't want to block on that!

jdavredbeard:
> in general, we don't really need the multi-language integration test pattern anymore, since the engine's language conformance testing is now very reliable, so the YAML test would be enough. But I don't want to block on that!

Good to know! Well those tests are currently skipped anyhow, so we could delete the unneeded tests once the TFE env vars get fixed

PR #1044, opened 2026-06-02
Fix CI - fail tests with missing credentials

### LINKED ISSUE #1021: Integration tests missing TFE credentials since ESC migration
## Summary

The integration tests in `examples/` that require TFE credentials (`TFE_ORGANIZATION`, `TFE_TOKEN`) have been silently skipped in CI since the ESC migration in #912 (Dec 2025).

## Root Cause

Before #912, `TFE_ORGANIZATION` and `TFE_TOKEN` were set in the workflow's global `env:` block, making them available to all jobs and steps. The ESC migration removed them from the global env and added them only to the `test_provider` step in the `prerequisites` job (via `renderLocalEnv` in the ci-mgmt template).

The `Run tests` step in the `test` job — which runs the integration tests in `examples/` — only gets `GITHUB_TOKEN`. As a result, `getRemoteBackendOrganization()` and `getRemoteBackendToken()` call `t.Skipf()`, and all remote backend tests are silently skipped.

## Evidence

- CI logs show the Go integration tests completing in ~22s with `coverage: [no statements]`
- With the TFE env vars manually added to the test step, integration tests run and pass: https://github.com/pulumi/pulumi-terraform/actions/runs/25390722167/job/74465106560

## Fix

The `Run tests` step in the workflow needs `TFE_ORGANIZATION` and `TFE_TOKEN` from ESC secrets. This may require an update to the ci-mgmt native template (`provider-ci/internal/pkg/templates/native/.github/workflows/run-acceptance-tests.yml`), which currently only injects `renderLocalEnv` into the `test_provider` step but not the `Run tests` step.

## Related

- #912 — ESC migration that introduced this regression
- `pulumi/ci-mgmt` — upstream template may need updating

iwahbe:
Fixes #1021

### DIFF
--- a/.github/workflows/run-acceptance-tests.yml
+++ b/.github/workflows/run-acceptance-tests.yml
@@ -422,6 +422,8 @@ jobs:
         cd examples && go test -count=1 -cover -timeout 2h -tags=${{ matrix.language }} -parallel 4 .
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        TFE_ORGANIZATION: ${{ steps.esc-secrets.outputs.TFE_ORGANIZATION }}
+        TFE_TOKEN: ${{ steps.esc-secrets.outputs.TFE_TOKEN }}
     - if: failure() && github.event_name == 'push'
       name: Notify Slack
       uses: 8398a7/action-slack@77eaa4f1c608a7d68b38af4e3f739dcd8cba273e # v3.19.0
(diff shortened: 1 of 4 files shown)


iwahbe:
From 6b017c4:

> The `test` job's `Run tests` step only received `GITHUB_TOKEN`, so
`TFE_ORGANIZATION` and `TFE_TOKEN` were unavailable to the `examples/`
integration tests. Combined with the preceding commit (which fails
rather than skips when these are missing in CI), the remote backend
tests fail.
> 
> The ci-mgmt native template hardcodes this step's env and only applies
`renderLocalEnv` to the prerequisites `Test Provider Library` step, so
there is no `.ci-mgmt.yaml` workaround. This is a manual edit to a
generated file as a stopgap to unblock CI; the durable fix belongs in
the ci-mgmt template and will supersede this on the next regeneration.

